PR #1, opened 2020-05-15
pqg quickfix

sgeisler:
A good enough for now fix for the pqg rendering problem :stuck_out_tongue_winking_eye: (seems to work in practice with our current font size).


vroland:
I think this is ok for now, since it will be obsolete with font backgrounds anyway.

PR #2, opened 2020-05-15
Clear screen if too many partial updates happened

sgeisler:
This PR adds a rough estimation how bad artifacts are obscuring the screen and clears/repaints it completely once a threshold is hit. Builds on #1. But could be easily separated if #1 is too hacky.

### DIFF
--- a/examples/terminal/main/config.h
+++ b/examples/terminal/main/config.h
@@ -2,6 +2,8 @@
 #include "firacode.h"
 #include "firacode_bold.h"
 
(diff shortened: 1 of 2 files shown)


vroland:
Looks like this eliminates regular full screen updates as well? I think it would still be useful to clear when switching to the alternate screen or doing a full terminal clear...

sgeisler:
It's just a heuristic. So if we get any information from the host to update the whole screen it would be good to do so. I saw some code doing something like that on some clear command afaik. Would I break it with these changes?

vroland:
I'll check that, then merge :)

Issue #4, opened 2020-05-26
BOM list


kennethlimcp:
Nice work @vroland ! I am trying to fabricate one board and get the dusty `ED060SC4` part working.

Are you able to generate the BOM so that it's easier to have the parts ordered? 

Or if you have a spare one, i wont mind buying it as well! My previous attempt to design a board that drinks the e-ink failed and your board will be useful to help me play around, test the e-ink displays to make sure it's working before making a revision based off your design. :)

vroland:
Hmm, yes, I should compile a definitive BOM at some point...
Until then, https://github.com/vroland/epdiy/blob/master/hardware/epaper-breakout/gerbers/bom_jlcpcb.csv is a start, but it is missing the USB connector and the ESP Module (hopefully that's it).

kennethlimcp:
Ah I missed that! Should be useful thanks!!

sebastius:
Also please update the PDFs, both are outdated :)

vroland:
As soon as I get the last parts of my new order, I'll solder a board which I have a complete and accurate BOM for which will mainly be compiled from LCSC and JLCPCB orders.

PR #5, opened 2020-05-28
Updated PDF and gerbers

sebastius:
Fixes _Originally posted by @sebastius in https://github.com/vroland/epdiy/issues/4#issuecomment-634985123_


vroland:
These gerbers are re-exported from the original v2? I think there is not really a need for that, since the current gerbers.zip is the one I ordered, so I would leave it as currently tested version.

sebastius:
Yes they are based on your V2, which i've built here at home. Feel free to drop the gerbers.

vroland:
removed gerbers in #7.

Issue #3, opened 2020-05-24
Terminal example 'crashes' USB Serial chip


sebastius:
Somehow the terminal example annoys the USB Serial chip so much it causes to re-enumerate on the USB bus :sunglasses: 

Tried it with the Arduino IDE serial terminal, screen and IDF monitor. All three are gloriously killed :+1: 

Could be a shitty driver on my Linux Mint, could be a bad chip. Haven't seen this behaviour before in other USB serial chips, this is my first board with the CH330.

sgeisler:
Interesting, when does that happen? I haven't actively experienced that so far, but maybe I just didn't notice? But my setup is a bit hacky too: let IDF monitor configure the serial port and then pipe the output of `script` to it to mirror a local shell for testing.

sebastius:
Oh it might be power related, it hangs my entire hub.

vroland:
Ha, that's interesting. So far I have not been able to reproduce it with neither of my machines... 
Which kernel version are you using? And it does not happen with the other examples?

sebastius:
It seems to be a problem when i'm doing both serial AND updating the eink. Maybe i ordered the wrong parts and have an overcurrent problem. Or a lack of capacitors on the main 5v and 3.3v rails. No clue yet.

It also breaks the USB connection in Windows. Its probably hardware :)

vroland:
Hm, jup, seems to be hardware. When I've got the time I can test it on a Raspberry Pi, which is not known for its particularly stable usb power, so I might be able to reproduce it there.

vroland:
Hm, I've tried it with my raspberry pi, still no unexpected behaviour :/

sebastius:
![image](https://user-images.githubusercontent.com/14947839/83074119-f0356080-a071-11ea-8a03-e57d99c14ce8.png)

My problem is definetly in the capacitor area (this is the 3.3v rail at the CH330N chip) . Will debug further!

sebastius:
More digging: this also happens on the 5v rail. Added a 10nF cap to the CH330N, no effect.

sebastius:
Quick question: the LT1945 seems to be powered of the 3.3v rail instead of the 5v rail. Why is that?

sebastius:
[...]

PR #6, opened 2020-05-28
Selection jumper 5v/3v, also gnd planes

sebastius:
This should solve #3 by making the voltage to the LT1945 selectable (default at 5v).

Redid some ground traces and planes for better ground coverage.


vroland:
Looks good so far, thanks for contributing :) 
However, my KiCad (5.1.6) complains about some via clearances (with its default settings, 0.25mm maybe?):
clearances
Could you look into it?

sebastius:
Should be fine: JLC has the following minimum: 5mil (0.127mm)

vroland:
Do you know if that's true for other manufacturers as well? I mean, we should not constrain us here unnecessarily. It seems most of them only need a little wiggle, so I could do it as well.

sebastius:
Feel free to wiggle :)

vroland:
Could you rebase this to remove  fd648a7 and 95a5599?

vroland:
Aand quickly rebase onto the current master? :)

sebastius:
I think i've done it, a friend helped me by telling me what to do ;)

vroland:
@sebastius Is there a specific reason you got rid of the mounting holes? I think it would be nice to have them, though I would be open to moving them to different locations.

sebastius:
I didn't remove the mounting holes on purpose, sorry! They should be there, if i removed them, we'll need to look into bringing them back, or reverting and redoing my mods. Which do you prefer?

vroland:
Ah, probably because they where not in the schematic, to KiCad removed them when updating the PCB. So I would re-add them and move them to their old positions. 
Do you want to do it (and re-position the ground vias) or shoud I do it at some time?

PR #10, opened 2020-05-28
unused opamp should be neutralized

sebastius:
Opamp 2 in the LM358 should be disabled to prevent damage to the chip and to reduce power consumption.
image
Sources: http://www.ti.com/lit/an/sboa204a/sboa204a.pdf https://www.analog.com/en/analog-dialogue/raqs/raq-issue-46.html


vroland:
Nice catch, didn't know about this one.

PR #13, opened 2020-06-03
Fixed value R14 to 390k as 400k doesnt exist and it will work anyway

sebastius:
See title. Boring little fix but it helps.


vroland:
:+1: I was using a 390k resistor already, anyway :D

PR #20, opened 2020-07-06
Weather example

mmib:
New example added. Shows current weather and forecast using openweathermap API.
New draw functions added.


vroland:
Wow, that looks great, thank you! :+1:   I'm currently waiting for my API key to get activated ;) 
I'll make a few remarks regarding the source code though, if that's alright with you.

components/epd_driver/epd_driver.c
```diff
@@ -301,39 +310,278 @@ inline uint32_t min(uint32_t x, uint32_t y) { return x < y ? x : y; }
...
+}
+
+void epd_fill_circle(int x0, int y0, int r, uint8_t color,
+                    uint8_t *framebuffer) {
+  epd_draw_vline(x0, y0 - r, 2 * r + 1, color, framebuffer);
+  epd_fill_circle_helper(x0, y0, r, 3, 0, color, framebuffer);
+}
+
+
+void epd_fill_circle_helper(int16_t x0, int16_t y0, int16_t r,
```
vroland, line 384: Is there a specific reason for the int16_t type? If not, I'd rather keep the types consistent and just use int.

components/epd_driver/epd_driver.c
```diff
@@ -301,39 +310,278 @@ inline uint32_t min(uint32_t x, uint32_t y) { return x < y ? x : y; }
...
+        epd_draw_vline(x0 + py, y0 - px, 2 * px + delta, color, framebuffer);
+      if (corners & 2)
+        epd_draw_vline(x0 - py, y0 - px, 2 * px + delta, color, framebuffer);
+      py = y;
+    }
+    px = x;
+  }
+}
+
+void epd_draw_rect(int16_t x, int16_t y, int16_t w, int16_t h,
```
vroland, line 427: Same.

examples/weather/main/weather.cpp
```diff
@@ -0,0 +1,1392 @@
...
+#include "sdkconfig.h"
+#include <stdio.h>
+#include <string.h>
+
+#include "epd_driver.h"
+
+#include "Arduino.h"
+#define ARDUINOJSON_ENABLE_ARDUINO_STRING 1
+#include <ArduinoJson.h>
+#include <HttpClient.h>
```
vroland, line 24: This library is called HTTPClient for me... Maybe a different version? (Using arduino-esp for IDF 4.0)

components/epd_driver/i2s_data_bus.c
```diff
@@ -162,8 +162,9 @@ void i2s_bus_init(i2s_bus_config *cfg) {
   // (Smallest possible divider according to the spec).
   dev->sample_rate_conf.tx_bck_div_num = 2;
 
-  // Initialize Audio Clock (APLL) for 80 Mhz.
-  rtc_clk_apll_enable(1, 0, 0, 8, 1);
```
vroland, line 166: As long as we do not know exactly why this workaround works, I'd rather want to keep it at the frequency specified in the data sheet, at least on the main branch.

components/epd_driver/epd_driver.c
```diff
@@ -301,39 +310,278 @@ inline uint32_t min(uint32_t x, uint32_t y) { return x < y ? x : y; }
...
     }
+    err -= dy;
+    if (err < 0) {
+      y0 += ystep;
+      err += dx;
+    }
+  }
+}
+
+void epd_draw_line(int16_t x0, int16_t y0, int16_t x1, int16_t y1,
```
vroland, line 482: Same thing about types.

components/epd_driver/epd_driver.c
mmib: yep, let's use int

examples/weather/main/weather.cpp
mmib: both are fine for me, changed

components/epd_driver/i2s_data_bus.c
mmib: True, added this into menuconfig as new display type.
This is only way to drive such display for me.

components/epd_driver/epd_driver.c
mmib: done

components/epd_driver/epd_driver.c
mmib: done

components/epd_driver/epd_driver.c
mmib: done

mmib:
Please check it again

vroland:
Thanks, looks good to me. Introducing a new display type sounds like a more or less elegant solution.

PR #25, opened 2020-09-01
complete designators in BOM

SimonSelg:
The BOM (`hardware/BOM.csv`) did not include designators for all parts. This PR adds the missing ones. I hope I got it right, I double-checked every part, but still, mistakes can happen.


vroland:
Looks good, thank you :+1:

PR #26, opened 2020-09-01
add lcsc part number to symbols/schema

SimonSelg:
This allows for automatic BOM generation for ordering with JCLPCB/LCSC where they already place some of the components on the PCB ([documentation on that](https://support.jlcpcb.com/article/84-how-to-generate-the-bom-and-centroid-file-from-kicad), [more documentation](https://dubiouscreations.com/2019/10/21/using-kicad-with-jlcpcb-assembly-service/)).

As a side note, the `epaper-breakout.sch-bak` should probably removed.


vroland:
> As a side note, the `epaper-breakout.sch-bak` should probably removed.

Yes, probably :)

Issue #40, opened 2020-11-27
README for terminal?


BackSlasher:
I just got a device to run the demo.
However, I can't get the terminal to run. I assumed that it's the same method of building and running, and then writing stuff to `/dev/ttyUSB0` (e.g. using `script -f /dev/ttyUSB0`), but no joy, reading from `UART_NUM_1` returns with 0.
Should I have built something differently?

I'm using the rev2/3 config

Thanks!

vroland:
Hm, that sounds like it should work. Is /dev/ttyUSB0 set up correctly? (115200 baud, no flow control)
You could also try to use different pins and hook up an external usb to serial device, so you still have serial output.

BackSlasher:
@vroland I'll try. Could you dump your tty setup if it's no bother? I'll try copying it :D

vroland:
I just use some serial monitor (e.g. minicom, idf monitor) to set it up ;) Maybe one thing you could try: uncomment the //esp_log_set_vprintf(log_to_uart); line.

BackSlasher:
If I uncomment the line it actually prevents anything from being printed to idf monitor.
Without it, I can still debug-print stuff and catch them on `idf monitor`, and see that `uart_read_bytes` always returns 0.
I see that idf monitor seems to set the TTY correctly (at least for baud)
```
$ stty -F /dev/ttyUSB0
speed 115200 baud; line = 0;
min = 0; time = 0;
-brkint -icrnl -imaxbel
-opost -onlcr
-isig -icanon -iexten -echo -echoe -echok -echoctl -echoke
```
Anything obvious I'm missing here?

vroland:
Reading through the code I found that the file in master actually uses pins GPIO_NUM_15, GPIO_NUM_14, which are broken out pins and not the RX / TX pins. Can you check which ones you are using?

BackSlasher:
I want to say "uhhhh the USB ones" but I know it's a bad answer :D
How easy is it to use the RX/TX ones, that I assume are already connected to the USB?
I'm not sure which layout the board I have is based on, would be happy to give you any number printed on the PCB if it helps

vroland:
Try GPIO_NUM_1 and GPIO_NUM_3 (don't know which one is which, sorry :D), these are the pins labeled TX/ RX on the module and those are the ones the ESP is flashed with. When using those, you will also have to uncomment the line described earlier, because the default serial output will be overridden.

BackSlasher:
```
[...]

PR #41, opened 2020-11-27
Terminal demo: Changed pins to USB TX/RX

BackSlasher:
Also added a bit of docs

Solves #40


BackSlasher:
I'll actually add a README, now that I think about it

BackSlasher:
done

vroland (approved):

PR #48, opened 2020-12-05
deinit function

mcer12:
Added epd_deinit function and small change in i2s_data_bus.c. i2s draws around 30uA extra in deep sleep, so I made a small change that mitigates that. Further improvements discussed in low power optimizations issue.


vroland:
Thank you, i didn't know that. I'll have a look later and then merge :)

vroland (approved):

PR #54, opened 2021-01-13
add Slack button in readme

mcer12:


vroland (approved):

vroland:
Thanks! The downside I see with slack is that people have to register to follow the discussion. But until we have something better this will do.

Issue #52, opened 2021-01-06
Noisy / grainy image with ED097TC2


2dom:
I just received my ED097TC2 panel to replace the broken ED097OC4. While it appears to be of better contrast it is also a lot more "grainy" (see right side of attached picture). Any suggestions what I can tweak to get rid of this?

Fun fact - if I place my hand on the display during update (placed it o the left side) the grain appears to be much reduced. 

![IMG_20210106_123041](https://user-images.githubusercontent.com/4075404/103764316-858ecf00-501b-11eb-85b9-0655804c10f8.jpg)

vroland:
You're right, when comparing the displays directly the TC2 seems noisier (though maybe not as much?) for me as well. I don't seem to have your magical hands though :/ Can you localize where your hand has an effect? I don't have an oscilloscope myself, but if I know what to look for I may be able to borrow one.
Is it the same if you first darken the display and then draw the image in white (The grayscale test example does both)?

2dom:
Will try white on black. I presume that the hand has some electrostatic effect. Maybe changes the off-set or something. I noticed this before - on another display I had put sum electrical tape on the back and it was clearly visible in the image.

vroland:
Hm, in my e-paper laptop I have the whole controller board taped to the back of the display with electrical tape but it's not visible in any way. Maybe because it is shielded? 
Which revision of the board are you using, btw?

mcer12:
@2dom just to be sure, did you adjust the voltages? According to datasheet it requires +27V (instead of usual +22V). And did you adjust VCOM according to the sticker?

2dom:
I did adapt vcom but not to 27V ... Will have to give this a go ☺️

2dom:
Thanks for the reminder - did not actually read the datasheet

2dom:
@vroland Checking the Revision I have to postpone until I take the picture frame apart next time

2dom:
@mcer12 ... Can you send me the datasheet please?

mcer12:
@2dom https://www.welectron.com/mediafiles/datasheets/waveshare/Waveshare-9.7inch-e-Paper-HAT-Datasheet.PDF

2dom:
Great - thanks!

vroland:
@2dom Did you already check if the voltage has any effect?

2dom:
No difference unfortunately

mcer12:
@2dom what version of the board are you using?

[...]

PR #61, opened 2021-01-28
Vendor Waveforms and Library Refactoring

vroland:
When this is done, it will extend the EPDIY component, such that it can use the vendor-provided waveform files.
This will be slower because of the limited memory bandwidth of the ESP32, but will produce better-quality images (#52).

Unfortunately, it's currently unclear wether the waveform files are copyrightable, so I will not distribute them in this repository for now. Some can be found through searching the web or dumping flash chips. Long therm, it would be great to generate our own waveforms.
Ideas about this issue are always welcome!


mcer12:
@vroland This is very nice! Maybe you can add links in the documentation?

vroland:
@mcer12 Yes, I'll do that. I don't think we're big enough for anyone to care and remove the linked files, so maintaining a link repository should be possible.

PR #62, opened 2021-01-31
Terminal text contrast fades after 1 hour 11 minutes

markbirss:
Overflow caused by using int32_t instead of int64_t

vrolandAPP  6:02 PM
Nice catch, that's a classic overflow right here! esp_timer_get_time() returns an int64_t, but t_poweron is uint32_t. So, after 2^32/10^6/60 = 71.5 min = 1 hour 11 minutes t_poweron overflows and time_since_poweron_m will be > 10. Changing the time of t_poweron to int64_t should fix it :)

### DIFF
--- a/examples/terminal/main/main.c
+++ b/examples/terminal/main/main.c
@@ -20,8 +20,13 @@
 #include "epd_driver.h"
 #include "st.h"
 
+/* USB */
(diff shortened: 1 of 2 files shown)


markbirss:
I have tested it, but someone else is welcome to test also

vroland:
Obsolete as of #63.

PR #64, opened 2021-01-31
we missed this one

markbirss:
somehow overlooked this one also needed

Merge pull request #63 from markbirss/master

Terminal text contrast fades after 1 hour 11 minutes


vroland (approved):

vroland:
Right, this one wouldn't really cause problems because it only stores the difference, but better use the correct type from the beginning.

PR #66, opened 2021-02-03
Adjust solderpaste after JLCPCB feedback

bartlammers:
After sending the V5 gerber files JLCPCB I got some feedback on the top solder paste layer for the stencil. It did not match the copper layer exactly.
The components in question are:
- Micro-USB connector (J2)
- ED060SC4 (J4)

I changed them to match the copper layer


vroland:
Good to know, thanks. I didn't order a stencil yet, so they probably ignored it.

PR #72, opened 2021-03-28
Add software rotation

martinberlin:
Points to cover are 3

1. epd_draw_pixel so all other geometrical functions work
2. Fonts (fonts.c)
3. Partial update (highlevel.c epd_hl_update_area)

On 3 is where I need still more testing. 1 & 2 seem to work like expected. 

**Added functions in epd_driver:**

```c
epd_set_rotation(0 to 3); // 0 is default as it was without rotation
epd_get_rotation();       // Will return the epaper rotation
```

### DIFF
--- a/examples/demo/main/main.c
+++ b/examples/demo/main/main.c
@@ -184,8 +184,7 @@ void idf_loop() {
   epd_poweroff();
 
   delay(5000);
-  epd_hl_set_all_white(&hl);
-
+  
   EpdRect board_area = {
       .x = epd_rotated_display_width() / 2 - img_board_width / 2,
       .y = epd_rotated_display_height() / 2 - img_board_height / 2,
@@ -249,8 +248,9 @@ void idf_setup() {
 
   epd_init(EPD_LUT_1K);
   hl = epd_hl_init(WAVEFORM);
-
-  epd_set_rotation(1);
+  //Optional but can be used directly
+  //enum EpdRotation display_rotation = EPD_ROT_INVERTED_LANDSCAPE;
+  epd_set_rotation(EPD_ROT_LANDSCAPE);
 
   printf("Dimensions after rotation, width: %d height: %d\n\n", epd_rotated_display_width(), epd_rotated_display_height());
 }
--- a/examples/demo/main/main.c
+++ b/examples/demo/main/main.c
@@ -161,7 +161,9 @@ void idf_loop() {
       .width = img_giraffe_width,
       .height = img_giraffe_height,
   };
-  epd_copy_to_framebuffer(giraffe_area, img_giraffe_data, fb);
+  //epd_copy_to_framebuffer(giraffe_area, img_giraffe_data, fb);
+  epd_draw_rotated_image(giraffe_area, img_giraffe_data, fb);
+
   epd_poweron();
   err = epd_hl_update_screen(&hl, MODE_GC16, temperature);
   epd_poweroff();
@@ -174,7 +176,8 @@ void idf_loop() {
       .width = img_zebra_width,
       .height = img_zebra_height,
   };
-  epd_copy_to_framebuffer(zebra_area, img_zebra_data, fb);
+  //epd_copy_to_framebuffer(zebra_area, img_zebra_data, fb);
+  epd_draw_rotated_image(zebra_area, img_zebra_data, fb);
   epd_poweron();
   err = epd_hl_update_screen(&hl, MODE_GC16, temperature);
   epd_poweroff();
@@ -187,7 +190,8 @@ void idf_loop() {
       .width = img_board_width,
       .height = img_board_height,
   };
-  epd_copy_to_framebuffer(board_area, img_board_data, fb);
+  //epd_copy_to_framebuffer(board_area, img_board_data, fb);
+  epd_draw_rotated_image(board_area, img_board_data, fb);
   cursor_x = epd_rotated_display_width() / 2;
   cursor_y = board_area.y;
   font_props.flags |= EPD_DRAW_BACKGROUND;
@@ -227,7 +231,8 @@ void idf_loop() {
       .width = img_beach_width,
       .height = img_beach_height,
   };
-  epd_copy_to_framebuffer(img_beach_area, img_beach_data, fb);
+  //epd_copy_to_framebuffer(img_beach_area, img_beach_data, fb);
+  epd_draw_rotated_image(img_beach_area, img_beach_data, fb);
 
   epd_poweron();
   err = epd_hl_update_screen(&hl, MODE_GC16, temperature);
@@ -244,7 +249,7 @@ void idf_setup() {
   epd_init(EPD_LUT_1K);
   hl = epd_hl_init(WAVEFORM);
 
-  epd_set_rotation(2);
+  epd_set_rotation(1);
 
   printf("Dimensions after rotation, width: %d height: %d\n\n", epd_rotated_display_width(), epd_rotated_display_height());
 }
--- a/examples/demo/main/main.c
+++ b/examples/demo/main/main.c
@@ -74,21 +74,28 @@ void draw_progress_bar(int x, int y, int width, int percent, uint8_t* fb) {
 void idf_loop() {
 
   enum EpdDrawError err;
+  
   temperature = epd_ambient_temperature();
   printf("current temperature: %d\n", temperature);
 
   uint8_t* fb = epd_hl_get_framebuffer(&hl);
+  
 
   epd_poweron();
   epd_clear();
   epd_poweroff();
 
-
-  int cursor_x = EPD_WIDTH / 2;
+  epd_fill_circle(30,30,15,0,fb);
+  int cursor_x = EPD_WIDTH / 2 - 200;
   int cursor_y = EPD_HEIGHT / 2 - 100;
   EpdFontProperties font_props = epd_font_properties_default();
   font_props.flags = EPD_DRAW_ALIGN_CENTER;
-  epd_write_string(&FONT, "Loading demo...", &cursor_x, &cursor_y, fb, &font_props);
+  font_props.fg_color = 120;
+
+  char srotation[32];
+  sprintf(srotation, "Loading demo...\nRotation: %d", epd_get_rotation());
+
+  epd_write_string(&FONT, srotation, &cursor_x, &cursor_y, fb, &font_props);
 
   int bar_x = EPD_WIDTH / 2 - 200;
   int bar_y = EPD_HEIGHT / 2;
@@ -109,7 +116,7 @@ void idf_loop() {
     vTaskDelay(1);
   }
 
-  cursor_x = EPD_WIDTH / 2;
+  cursor_x = EPD_WIDTH / 2 - 200;
   cursor_y = EPD_HEIGHT / 2 + 200;
 
   epd_write_string(&FONT, "Just kidding,\n this is a demo animation 😉", &cursor_x, &cursor_y, fb, &font_props);
@@ -129,6 +136,7 @@ void idf_loop() {
   cursor_x = EPD_WIDTH / 2;
 
   delay(1000);
+  return;
 
   EpdRect clear_area = {
     .x = 0,
@@ -244,8 +252,8 @@ void idf_setup() {
 void app_main() {
   idf_setup();
 
-  while (1) {
+  //while (1) {
     idf_loop();
-  };
+  //};
 }
 #endif
--- a/examples/demo/main/main.c
+++ b/examples/demo/main/main.c
@@ -116,7 +116,7 @@ void idf_loop() {
   }
 
   cursor_x = epd_rotated_display_width() / 2;
-  cursor_y = epd_rotated_display_height() / 2 + 200;
+  cursor_y = epd_rotated_display_height() / 2 + 100;
 
   epd_write_string(&FONT, "Just kidding,\n this is a demo animation 😉", &cursor_x, &cursor_y, fb, &font_props);
   epd_poweron();
@@ -169,6 +169,7 @@ void idf_loop() {
   epd_poweroff();
 
   delay(5000);
+  epd_hl_set_all_white(&hl);
 
   EpdRect zebra_area = {
       .x = epd_rotated_display_width() / 2 - img_zebra_width / 2,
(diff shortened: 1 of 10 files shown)


src/epd_driver/epd_driver.c
```diff
@@ -6,6 +6,8 @@
 #include "esp_log.h"
 #include "esp_types.h"
 
+// Initial rotation: Landscape
+uint8_t display_rotation = 0;
```
vroland, line 10: This variable should be static, since it does not need to be exposed to the outside.
Also, an enum type for the display orientation would be nicer?

src/epd_driver/epd_driver.c
```diff
@@ -40,13 +41,73 @@ void epd_draw_vline(int x, int y, int length, uint8_t color,
...
 }
 
+EpdRect _rotated_area(uint16_t x, uint16_t y, uint16_t w, uint16_t h) {
+  // If partial update uses full screen do not rotate anything
+  if (x != 0 && y != 0 && EPD_WIDTH != w && EPD_HEIGHT != h) {
+    switch (epd_get_rotation())
+    {
+      // 0 landscape: Leave it as is
+      // 1 90 ° right portrait
+      case 1:
```
vroland, line 51: Same here, using enums for the orientation would be nicer, since otherwise there are these magic numbers and you have to remember which is which.

src/epd_driver/epd_driver.c
```diff
@@ -40,13 +41,73 @@ void epd_draw_vline(int x, int y, int length, uint8_t color,
...
+    return coord;
+}
+
 void epd_draw_pixel(int x, int y, uint8_t color, uint8_t *framebuffer) {
-  if (x < 0 || x >= EPD_WIDTH) {
+  if (x < 0 || x >= epd_rotated_display_width()) {
     return;
   }
-  if (y < 0 || y >= EPD_HEIGHT) {
+  if (y < 0 || y >= epd_rotated_display_height()) {
```
vroland, line 103: If I'm not mistaken, these checks should be moved to after the rotation and then check for EPD_WIDTH / HEIGHT as is. This might be equivalent, but to me more confusing since these checks are meant to guard for the actual buffer bounds.

src/epd_driver/include/epd_driver.h
```diff
@@ -25,6 +25,13 @@ typedef struct {
   int height;
 } EpdRect;
 
+// Simple x and y coordinate
```
vroland, line 28: Since this type is only used internally, it should remain in the C file and not be publicly exposed.

src/epd_driver/include/epd_driver.h
```diff
@@ -224,6 +243,16 @@ EpdRect epd_full_screen();
 void epd_copy_to_framebuffer(EpdRect image_area, const uint8_t *image_data,
                              uint8_t *framebuffer);
 
+/**
```
vroland, line 246: Same for these functions. They are not public, so they should not appear in the public header.

src/epd_driver/epd_driver.c
martinberlin: Sure, can you suggest the right ones? 
The cool thing with an int is that it can be validated just doing <4. This is also the GxEPD way to handle rotation so many people is quite familiar with it.

src/epd_driver/epd_driver.c
martinberlin: I agree. But is just Nx90 clockwise, is not so hard :)

src/epd_driver/epd_driver.c
vroland: EPD_ROT_LANDSCAPE, EPD_ROT_PORTRAIT, EPD_ROT_INVERTED_LANDSCAPE and EPD_ROT_INVERTED_PORTRAIT?
And in C, enums are really just int ;) So you can just give them the values 0 - 3 and the could would just work when using ints. Enums are just nicer because they hold more semantic information.

src/epd_driver/epd_driver.c
vroland: True, but you only have to read the code once with enums to understand it, while you might have to think for 5 seconds more without them ;)

src/epd_driver/include/epd_driver.h
```diff
@@ -13,6 +13,14 @@ extern "C" {
 #include <stdbool.h>
 #include <stdint.h>
 
+/// Display software rotation. Recalculates x, y of draw_pixel, fonts and partial update area
```
vroland, line 16: Maybe elaborate on the documentation, like mentioning epd_set_rotation for setting the rotation and that it sets the rotation for the purposes of the drawing and font functions. I think "recalculate" sounds a bit misleading.

src/epd_driver/include/epd_driver.h
```diff
@@ -177,6 +185,18 @@ typedef struct {
 /** Initialize the ePaper display */
 void epd_init(enum EpdInitOptions options);
 
+/** Rotation getter */
+uint8_t epd_get_rotation();
+
+/** Rotation setter */
```
vroland, line 191: Same here, rotation getter and setter doesn't explain very much.

src/epd_driver/epd_driver.c
```diff
@@ -40,13 +47,39 @@ void epd_draw_vline(int x, int y, int length, uint8_t color,
   }
 }
 
+Coord_xy _rotate(uint16_t x, uint16_t y) {
+    switch (display_rotation) {
+        case 1:
```
vroland, line 52: Use the enums instead of numbers.

src/epd_driver/epd_driver.c
```diff
@@ -342,3 +375,80 @@ enum EpdDrawError epd_draw_image(EpdRect area, const uint8_t *data, const EpdWav
...
+ } else {
+   buf_val = (buf_val & 0x0F);
+ }
+
+ // epd_draw_pixel needs a 0 -> 255 value
+ return buf_val<<4;
+}
+
+void epd_draw_rotated_image(EpdRect image_area, const uint8_t *image_buffer, uint8_t *framebuffer) {
+  if (epd_get_rotation()!=0) {
```
vroland, line 435: Please use enums instead of the number for the rotation.

vroland (changes requested):

src/epd_driver/include/epd_driver.h
```diff
@@ -460,8 +480,15 @@ EpdRect epd_difference_image_cropped(
  */
 EpdRect epd_difference_image(const uint8_t* to, const uint8_t* from, uint8_t* interlaced, bool* dirty_lines);
 
+/**
+ * Return the pixel color of a 4 bit image array
+ */
+uint8_t epd_get_pixel(int x, int y, int fb_width, int fb_height, const uint8_t *framebuffer);
```
vroland, line 486: Please add a more detailed description of the parameters, like that the color is 0-255 as in epd_draw_pixels.

vroland (changes requested):

martinberlin (commented):
Moved this out, can you please check if I understood well?

examples/grayscale_test/main/main.c
```diff
@@ -55,17 +55,19 @@ void loop() {
```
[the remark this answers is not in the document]
martinberlin: Done here: https://github.com/martinberlin/epdiy-rotation/commit/c3b10cf7334ab83e91c6b6db4eb2cdbf923429bb

examples/demo/main/main.c
```diff
@@ -236,14 +248,19 @@ void idf_setup() {
...
+  epd_set_rotation(EPD_ROT_LANDSCAPE);
+
+  printf("Dimensions after rotation, width: %d height: %d\n\n", epd_rotated_display_width(), epd_rotated_display_height());
 }
 
 #ifndef ARDUINO_ARCH_ESP32
 void app_main() {
   idf_setup();
 
-  while (1) {
```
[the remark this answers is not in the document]
martinberlin: Left the loop like originally was:
https://github.com/martinberlin/epdiy-rotation/commit/c3b10cf7334ab83e91c6b6db4eb2cdbf923429bb

src/epd_driver/epd_driver.c
```diff
@@ -6,6 +6,14 @@
...
 #include "esp_types.h"
 
+// Simple x and y coordinate
+typedef struct {
+    uint16_t x;
+    uint16_t y;
+} Coord_xy;
+
+// Initial rotation: EPD_ROT_LANDSCAPE
+static uint8_t display_rotation = 0;
```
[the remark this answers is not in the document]
martinberlin: Thanks! Updated here: https://github.com/martinberlin/epdiy-rotation/commit/10c02f95060404e3a3dcae74bb372200464a3e19

src/epd_driver/epd_driver.c
martinberlin: Good catch. Part in last commit, this last two here: https://github.com/martinberlin/epdiy-rotation/commit/94964b48b1e4450d564ae5d5da1de066a68fa494

src/epd_driver/font.c
```diff
@@ -171,11 +174,9 @@ static enum EpdDrawError IRAM_ATTR draw_char(const EpdFont *font, uint8_t *buffe
         bm = bm >> 4;
       }
 
-      if ((xx & 1) == 0) {
-        buffer[buf_pos] = (old & 0xF0) | color_lut[bm];
-      } else {
-        buffer[buf_pos] = (old & 0x0F) | (color_lut[bm] << 4);
-      }
+      color = (old & 0x0F) | (color_lut[bm] << 4);
```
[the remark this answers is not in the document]
martinberlin: Thanks for the review. Updated here: https://github.com/martinberlin/epdiy-rotation/commit/2903da43d3f0b841baac98443c809c7538c52a94

src/epd_driver/highlevel.c
```diff
@@ -45,12 +54,53 @@ enum EpdDrawError epd_hl_update_screen(EpdiyHighlevelState* state, enum EpdDrawM
...
 }
 
+EpdRect _rotated_area(uint16_t x, uint16_t y, uint16_t w, uint16_t h) {
+  // If partial update uses full screen do not rotate anything
+  if (x != 0 && y != 0 && EPD_WIDTH != w && EPD_HEIGHT != h) {
+    switch (epd_get_rotation())
+    {
+      // 0 landscape: Leave it as is
+      // 1 90 ° right portrait
+      case 1:
```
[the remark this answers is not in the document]
martinberlin: Good catch. Updated in [last commit](https://github.com/martinberlin/epdiy-rotation/commit/2903da43d3f0b841baac98443c809c7538c52a94)

src/epd_driver/include/epd_driver.h
martinberlin: Please check if [this is better](https://github.com/martinberlin/epdiy-rotation/commit/1bbc2d477bf8d7447968f87218b394d401945459)

src/epd_driver/include/epd_driver.h
martinberlin: Check in [last commit](https://github.com/martinberlin/epdiy-rotation/commit/1bbc2d477bf8d7447968f87218b394d401945459) if the detailed description is better suited

src/epd_driver/epd_driver.c
martinberlin: This is the only point that I'm not sure how to solve. Can you specify better how would you do it?

martinberlin (commented):
Please check if the review is covering the points you need. Only one point that I still don't know what is the best solution is the bounds check in draw_pixel. This one you can also modify later since seems an easy one and that way you can add it in the best way you imagine.
Thanks a lot for the review!

src/epd_driver/epd_driver.c
```diff
@@ -342,3 +377,84 @@ enum EpdDrawError epd_draw_image(EpdRect area, const uint8_t *data, const EpdWav
     };
     return epd_draw_base(area, data, no_crop, EPD_MODE_DEFAULT, temperature, NULL, waveform);
 }
+
+void epd_set_rotation(enum EpdRotation rotation) {
+      display_rotation = rotation;
+}
+
+uint8_t epd_get_rotation() {
```
vroland, line 385: This should also return enum EpdRotation.

src/epd_driver/epd_driver.c
```diff
@@ -342,3 +377,84 @@ enum EpdDrawError epd_draw_image(EpdRect area, const uint8_t *data, const EpdWav
...
+    return display_rotation;
+}
+
+uint16_t epd_rotated_display_width() {
```
vroland, line 389: I'd just use int here, since C converts to int for calculations anyway and on the esp32 a 16 bit int is not more efficient than a 32 bit int.

src/epd_driver/epd_driver.c
```diff
@@ -342,3 +377,84 @@ enum EpdDrawError epd_draw_image(EpdRect area, const uint8_t *data, const EpdWav
...
+              display_width = EPD_HEIGHT;
+          break;
+          case EPD_ROT_INVERTED_LANDSCAPE:
+          case EPD_ROT_LANDSCAPE:
+          break;
+      }
+  return display_width;
+}
+
+uint16_t epd_rotated_display_height() {
```
vroland, line 405: uint16_t -> int

vroland (changes requested):
Just two things about the types, otherwise this is looking pretty good now :)

src/epd_driver/epd_driver.c
martinberlin: Updated in [last commit](https://github.com/martinberlin/epdiy-rotation/commit/bde584b9dd0dd0eb401f6a845eb9cafcdda8d744). Thanks for the detailed and clear review after my update!

src/epd_driver/epd_driver.c
martinberlin: Updated in [last commit](https://github.com/martinberlin/epdiy-rotation/commit/bde584b9dd0dd0eb401f6a845eb9cafcdda8d744)

martinberlin (commented):
Thanks for the great review. A pleasure to add this function and at the same time to learn more of how your component is organized

vroland (approved):

vroland:
Looks good to me! Thank you very much for implementing this and patiently incorporating the reviews :)

Issue #12, opened 2020-06-02
Use the terminal example as an actual terminal


sebastius:
I had great fun already using the Serial Terminal example, and with a bit of foolishness i added a second USB serial to route my PC keyboard from one terminal into the e-ink terminal session (by connecting the second USB-Serial TX line to the RX of the CH330). One problem arose when trying to update the board: the second usb-serial causes the communication to fail. 

I'd like to have a proper input method on the ESP32. Could be as simple as just using a second serial port (they can be routed to any GPIO) and then pushing them towards the regular serial output. That will also allow some control options, like the classical terminals used to have. Alternatively a PS2 port, Bluetooth host or USB host could also work. I don't think we have the GPIOs left for a full keyboard matrix (i think you'd need about 20 GPIOs to do it).

Any thoughts on what you'd like to do in terms of input devices? We could make the expansion header a bit more useful by adding in as much of the GPIOs as we can. 

(besides this hardware/software upgrade the terminal code would need a pretty decent upgrade too, to get to 80columns width, work with a cursor, stuff like VIM would be useful to have. But the first tests were really pleasing!)

Heck at some point i'd love to see the ESP32 do an SSH into my desktop :P (or maybe use a seperate ESP32 for that, so the current chip can focus on being a terminal).

vroland:
Yes, a lot of those things are planned for the future ;) First of all, to test how it looks with real terminal input, just run `script -f /dev/ttyUSB0` in a terminal emulator window (of approximately the size of the display, to set $COLUMNS and $ROWS correctly). Additionally, you'll need to open a serial monitor to empty the output buffer of the ESP.
This lets you try all the fun stuff, like:
![IMG_20200602_183810](https://user-images.githubusercontent.com/5757380/83546174-92d16180-a500-11ea-99ad-fd78997de23f.jpg)
I am also working on an AVR-based board to read keyboard and trackpad input from an old ThinkPad keyboard, which I'll write about once I've got all the parts and can assemble  it.
This microntroller can then talk with the ESP and send keyboard input via UART or SPI. 
So basically, once you connect an external microcontroller via the unused and broken out pins, you should be able to use pretty much anything ;)

Running ssh directly on the ESP is a sweet idea as well, maybe I'll come around to trying it at some point...

vroland:
[...]

PR #79, opened 2021-04-21
Added LilyGo T5 4.7inch EPD PlatformIO example

DavidM42:
This adds an example how to use this library with the LilyGo T5 4.7 inch EPD device with an Esp32 on board and some exra features like buttons on the side.
It includes
* A full PlatformIO config
* Some additional example code how to combine the epd-driver with the deepsleep functionalities of the Esp32
* An experimental (not ideal) calculation method to find out the battery percentage on this specific device

As discussed [here over on the repo of the manufacturer driver fork](https://github.com/Xinyuan-LilyGO/LilyGo-EPD47/issues/9#issuecomment-823364297) I think this would be a good kickstart for buyers of that device.

If anyone can test it and maybe suggest improvements to stop random screen greying of the background or how to improve the battery percentage calculation that would be greatly appreciated.

Thanks again for this nice library


vroland:
That looks interesting, thank you :) I'll try to review and test this this weekend!

fdev31:
I wanted to do something a bit similar, but it went a bit too far I guess... 
https://github.com/fdev31/LilygoT5-weather-photo-display

Also I'm not exactly certain about the best practices here... but It looks like you have more advanced battery handling code (mine was a simple example I borrowed somewhere). Is a lookup table really needed in theory ? (I thought some linear math as you did was good enough....)

examples/lilygo-t5-47-epd-platformio/platformio.ini
```diff
@@ -0,0 +1,39 @@
...
+;   Upload options: custom upload port, speed and extra flags
+;   Library options: dependencies, extra library storages
+;   Advanced options: extra scripting
+;
+; Please visit documentation for the other options and examples
+; https://docs.platformio.org/page/projectconf.html
+
+[common_env_data]
+framework = arduino
+board_build.f_cpu = 80000000L
```
vroland, line 13: I'm usually using 240 MHz, but I'm not sure if taking longer to process or running at a slower speed outweighs the other here energy-wise.

examples/lilygo-t5-47-epd-platformio/src/main.cpp
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * Upper most button on side of device. Used to setup as wakeup source to start from deepsleep.
+ * Pinout here https://ae01.alicdn.com/kf/H133488d889bd4dd4942fbc1415e0deb1j.jpg
+ */
+gpio_num_t FIRST_BTN_PIN = GPIO_NUM_39;
+
+EpdiyHighlevelState hl;
+// epd_ambient_temperature() exists
+// but did not work for this device in my testing
+// so hardcode to about your room temp
```
vroland, line 34: That's because the LilyGo boards don't have a temperature sensor.

examples/lilygo-t5-47-epd-platformio/src/main.cpp
```diff
@@ -0,0 +1,243 @@
...
+    return percent_experiment;
+}
+
+void display_center_message(const char* text) {
+    // first clear screen
+    epd_poweron();
+    epd_clear();
+    err = epd_hl_update_screen(&hl, MODE_GL16, temperature);
+    epd_poweroff();
+    delay(100);
```
vroland, line 96: For me, this produces rendering artifacts, because clear() is used without clearing the framebuffer.
I would clear the screen after power on, and then use the highlevel API:
`
void display_center_message(const char* text) {

    epd_hl_set_all_white(&hl);

    int cursor_x = EPD_WIDTH / 2;
    int cursor_y = EPD_HEIGHT / 2;
    if (orientation == EPD_ROT_PORTRAIT) {
        // height and width switched here because portrait mode
        cursor_x = EPD_HEIGHT / 2;
        cursor_y = EPD_WIDTH / 2;
    }
    EpdFontProperties font_props = epd_font_properties_default();
    font_props.flags = EPD_DRAW_ALIGN_CENTER;
    epd_write_string(&FiraSans_12, text, &cursor_x, &cursor_y, fb, &font_props);

    epd_poweron();
    err = epd_hl_update_screen(&hl, MODE_GC16, temperature);
    delay(500);
    epd_poweroff();
    delay(1000);
}
`
Also, the delays seem unnecessary for me?

examples/lilygo-t5-47-epd-platformio/src/main.cpp
```diff
@@ -0,0 +1,243 @@
...
+    Serial.begin(115200);
+
+    correct_adc_reference();
+
+    // First setup epd to use later
+    epd_init(EPD_OPTIONS_DEFAULT);
+    hl = epd_hl_init(WAVEFORM);
+    epd_set_rotation(orientation);
+    fb = epd_hl_get_framebuffer(&hl);
+
```
vroland, line 211: Maybe just clear the screen in the beginning.

vroland (changes requested):

vroland:
Thanks for the example, so far everything works with platformio. I noticed one issue where the text drawn after 30 seconds of run time produces weird artefacts. This is because the screen is cleared without setting the framebuffer to white. Could you fix that before the merge?

DavidM42:
Thanks for the great input. I'm new to electronic paper displays, so I'm thankful for all tips.
I adapted most of your requests, but I still sometimes encounter artefacts. If you have time maybe you can have a look at the reasons for that

examples/lilygo-t5-47-epd-platformio/platformio.ini
DavidM42: I just set it to 240 MHz too now. I don't think it has a substantial impact on energy consumption.
Honestly I just use 80 MHz because it's default in many examples and I never have problems with slow speeds.
For this epd driving it could be beneficial

examples/lilygo-t5-47-epd-platformio/src/main.cpp
DavidM42: I now use the `epd_hl_set_all_white` method, but the delays reduce some artifacts or graying of the background in my testing.
Maybe they can be avoided with further changes to the driving code here

vroland:
Thanks, looks good :)
Yes, there is still a little bit of aliasing after updating the font. This is because I use an approximate waveform, not the matching vendor waveform. With the vendor waveform it looks better, but LilyGO did not publish it in their repository and don't reply to the corresponding issue. I won't publish it myself, because it's unclear if these are copyright-able, and I don't want this repo to be taken down because of it.

vroland (approved):

entozoon:
@vroland LilyGo appear to have responded to you and open-sourced their [waveform](https://github.com/Xinyuan-LilyGO/LilyGo-EPD47/blob/master/Waveform_header/ED047TC2.h), which is awesome!
I've created a [PR](https://github.com/vroland/epdiy/pull/129) if you'd to merge. It's tested and looks great

@DavidM42 Thank you so much for this! Saved me so much time, I can't even.

DavidM42:
@entozoon I wanted to comment on your PR that I tested it and that it really looks great and sharp with the added waveform :tada: but it got merged already so here

I'm glad I could help you and my long struggle with this was less of a waste

entozoon:
Thanks @DavidM42, and yeah I'd have given up without your work. In fact, I'm [pretty close again](https://github.com/vroland/epdiy/issues/12#issuecomment-983142030) .. lol

PR #80, opened 2021-04-23
Text background now requires EPD_DRAW_BACKGROUND

fdev31:
Now text don't have "bounding box" backgrounds painted if the EPD_DRAW_BACKGROUND flag isn't set.
Allows painting over complex images (non flat color) without visual artifacts introduction.

### DIFF
--- a/src/epd_driver/font.c
+++ b/src/epd_driver/font.c
@@ -151,6 +151,7 @@ static enum EpdDrawError IRAM_ATTR draw_char(const EpdFont *font, uint8_t *buffe
     int color_difference = (int)props->fg_color - (int)props->bg_color;
     color_lut[c] = max(0, min(15, props->bg_color + c * color_difference / 15));
   }
+  bool background_needed = props->flags & EPD_DRAW_BACKGROUND;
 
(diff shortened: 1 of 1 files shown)


src/epd_driver/font.c
```diff
@@ -151,6 +151,7 @@ static enum EpdDrawError IRAM_ATTR draw_char(const EpdFont *font, uint8_t *buffe
     int color_difference = (int)props->fg_color - (int)props->bg_color;
     color_lut[c] = max(0, min(15, props->bg_color + c * color_difference / 15));
   }
+  char background_needed = props->flags & EPD_DRAW_BACKGROUND;
```
vroland, line 154: I'd use the bool type here for readability. Since the esp is a 32 bit system, there is no reason to prefer char.

vroland (changes requested):

vroland:
Looks good! I'd just do one small change: Using bool instead of char for the flag would be more readable and is irrelevant for performance. Could you change that?

fdev31:
Just did it ;)

vroland:
Thanks a lot!

PR #97, opened 2021-07-28
Add log error in red if PSRAM is not enabled with menuconfig

martinberlin:
Very easy 3 Liner to show a straightforward error when PSRAM is not enabled:

Screenshot from Ubuntu


vroland:
I think a LOGW would be better, since you only need PSRAM for the highlevel api. Not that anyone would want to do it, but hey ;)

martinberlin:
No problem. Updated!

vroland:
Looks good, thanks!

PR #99, opened 2021-08-16
Add JPG download and render example with SSL support

martinberlin:
Hello Valentin,
Please feel free to try it out and make it better or suggest updates. Tracked in [Issue 2](https://github.com/martinberlin/epdiy-rotation/issues/2) of my fork.

Prerequisites:

    This defines must be filled

```c
// www URL of the bitmap image (Should be BMP format 1 or 24 bits-depth)
#define IMG_URL "http://"
// WiFi configuration
#define ESP_WIFI_SSID     "WLAN-724300"
#define ESP_WIFI_PASSWORD ""
```


vroland:
Looking forward to review this soon, I think I can take the time on the weekend :)

martinberlin:
No rush. I also need to find time to do a cleanup since I adapted this to EPDiy from a download example of my component CalEPD and it still has some variables that are there doing nothing. So around friday night and saturday I will push a new update after testing. So far runs good in 1-bit which is nice for fast testing with a faster download and with 24-bit using the new formula.
Also feel free to prepare a 24-bit BMP that you like in the same resolution as the Dragon so we can upload it anywhere you like.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+#include <string.h>
+#include <math.h> // round + pow
+
+#include "epd_driver.h"
+#include "epd_highlevel.h"
+EpdiyHighlevelState hl;
+
+// WiFi configuration
+#define ESP_WIFI_SSID     "WLAN-724300"
+#define ESP_WIFI_PASSWORD "50238634630558382093"
```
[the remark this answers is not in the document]
vroland: If not, you should change this and squash the commits. Though it's kind of too late now.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+
+
+// Image URL and jpg settings. Make sure to update WIDTH/HEIGHT if using loremflickr
+// Note: Only HTTP protocol supported (Check README to use SSL secure URLs)
+#define IMG_URL "https://loremflickr.com/960/540"
```
vroland, line 45: I'd add something like this:
`
#define STR_HELPER(x) #x
#define STR(x) STR_HELPER(x)

#define IMG_URL ("https://loremflickr.com/"STR(EPD_WIDTH)"/"STR(EPD_HEIGHT))
`
So the images are in the right resolution. One less thing to change for the user.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+
+// Jpeg: Adds dithering to image rendering (Makes grayscale smoother on transitions)
+#define JPG_DITHERING true
+// Affects the gamma to calculate gray (lower is darker/higher contrast)
+// Nice test values: 0.9 1.2 1.4 higher and is too bright
+double gamma_value = 0.9;
+
+// As default is 512 without setting buffer_size property in esp_http_client_config_t
+#define HTTP_RECEIVE_BUFFER_SIZE 1536
+extern const uint8_t server_cert_pem_start[] asm("_binary_server_cert_pem_start");
```
vroland, line 56: Add an additional comment explaining this? Or just link to the documentation.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+uint8_t gamme_curve[256];
+
+static const char *TAG = "EPDiy";
+uint16_t countDataEventCalls = 0;
+uint32_t countDataBytes = 0;
+uint32_t img_buf_pos = 0;
+uint32_t dataLenTotal = 0;
+uint64_t startTime = 0;
+
+/*====================================================================================
```
vroland, line 103: Extend this notice?

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+
+  Created by Bodmer 15th Jan 2017
+  ==================================================================================*/
+
+// Return the minimum of two values a and b
+#define minimum(a,b)     (((a) < (b)) ? (a) : (b))
+
+uint8_t find_closest_palette_color(uint8_t oldpixel)
+{
+  return (round((oldpixel / 16)*16));
```
vroland, line 114: This is equivalent to `oldpixel & 0xF0`, right?

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+    for (uint32_t bx=0; bx<width;bx++) {
+        epd_draw_pixel(bx + padding_x, by + padding_y, decoded_image[by * width + bx], fb);
+    }
+  }
+  // calculate how long it took to draw the image
+  time_render = (esp_timer_get_time() - drawTime)/1000;
+  ESP_LOGI("render", "%d ms - jpeg draw", time_render);
+}
+
+void deepsleep(){
```
vroland, line 163: add `epd_deinit()`

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+               uint8_t *buff, // Pointer to the read buffer (NULL:skip) 
+               uint32_t nd 
+) {
+    vTaskDelay(1);
+    esp_task_wdt_reset();
+    uint32_t count = 0;
+
+    while (count < nd) {
+      if (buff != NULL) {
+            uint8_t b = source_buf[buffer_pos];
```
vroland, line 177: We could make source_buf much smaller by decompressing while the data comes in, but we'd need to keep track of how much was read / written. Are you interested in adding this?

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+        ++countDataEventCalls;
+        #if DEBUG_VERBOSE
+          if (countDataEventCalls%10==0) {
+            ESP_LOGI(TAG, "%d len:%d\n", countDataEventCalls, evt->data_len); 
+          }
+        #endif
+        dataLenTotal += evt->data_len;
+        
+        // Copy the response into the buffer
+        memcpy(output_buffer, evt->data, evt->data_len);
```
vroland, line 288: Add a check that evt->data_len does not overflow output_buffer.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    /* The event will not be processed after unregister */
+    ESP_ERROR_CHECK(esp_event_handler_instance_unregister(IP_EVENT, IP_EVENT_STA_GOT_IP, instance_got_ip));
+    ESP_ERROR_CHECK(esp_event_handler_instance_unregister(WIFI_EVENT, ESP_EVENT_ANY_ID, instance_any_id));
+    vEventGroupDelete(s_wifi_event_group);
+}
+
+void app_main() {
+  epd_init(EPD_OPTIONS_DEFAULT);
```
vroland, line 479: I used `epd_init(EPD_LUT_64K | EPD_FEED_QUEUE_8)`. The impact on rendering performance is pretty small, and then no additional configuring of mbedtls is needed, because enough byte-addressable IRAM is left.

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,518 @@
...
+      ret = nvs_flash_init();
+  }
+  ESP_ERROR_CHECK(ret);
+
+  // WiFi log level set only to Error otherwise outputs too much
+  esp_log_level_set("wifi", ESP_LOG_ERROR);
+  
+  // Initialization: WiFi + clean screen while downloading (optional)
+  wifi_init_sta();
+  epd_poweron();
```
vroland, line 514: The display should be cleared, because the highlevel API will forget the screen state after deepsleep (The PSRAM is powered down). To get accurate grayscale, the screen must be brought to a known state. If using the highlevel API, use `epd_fullclear(&hl, temp)`.

Also, we only powering on before draw (and clear) will waste less power.

examples/www-image/main/bmp-render.c
```diff
@@ -0,0 +1,592 @@
+/* Simple firmware for a ESP32 displaying a static image on an EPaper Screen */
```
vroland, line 1: This seems not functional ATM. I didn't see any epd_poweron()? 
Do you still want to include the bitmap version? It may be easier to maintain just the jpg vesion as an example.

vroland (changes requested):

martinberlin:
About BMP version:
>This seems not functional ATM. I didn't see any epd_poweron()?

You are right. But it refreshes, so if I'm not doing the poweron is being called automatically, or it does work even without invoking poweron()?

For me the BMP is a valid example. Since it does not need decompression and also supports 1-bit mode, that it can be used for some artistic reasons (In my point of view)

I cannot comment, I just can mark the conversation "as resolved", no idea why....

martinberlin:
>#define IMG_URL ("https://loremflickr.com/"STR(EPD_WIDTH)"/"STR(EPD_HEIGHT))
`
>So the images are in the right resolution. One less thing to change for the user.

Great idea. Added! But I would like to mention that I've never, so far, made the HTTPS request version. My mbedTLS settings look like:
Screenshot from 2021-08-22 15-52-48

But it does not matter if I select Internal Ram or external, in the LILYGO EPD47 it always hangs with HTTPS. With panic alert or some weird TLS error in the Certificate handshake. My question is: Can also try it in the EPD 47?
Maybe it does work with EPDiy v6 but not on the Lilygo?

Either that or you are using some special setting that I never catched. Either the way I feel insecure leaving the HTTPS version as default like this.

martinberlin (commented):
Pending to be resolved:

- [ ] Check if it’s possible to decompress src_buffer while downloading 
- [ ] Take a decision if to leave or remove BMP example
- [x] See if it’s possible to make the https request with the certificate

Rest of the review is covered if I did not skip anything. What a pity I cannot comment on your feedback on my last PR I could do that.

examples/www-image/main/jpg-render.c
martinberlin: Done, check my comments in the conversation below, I think it’s not really necessary since the buffer_size http configuration should already cover this.

examples/www-image/main/jpg-render.c
martinberlin: Done

examples/www-image/main/jpg-render.c
martinberlin: Corrected, thanks!

examples/www-image/main/jpg-render.c
martinberlin: I don’t know if I really got what had to be done here but I extended the comment leaving the credits to abdomen

examples/www-image/main/jpg-render.c
martinberlin: Added, thanks!

examples/www-image/main/jpg-render.c
martinberlin: Great advice, thanks!

examples/www-image/main/jpg-render.c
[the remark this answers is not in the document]
martinberlin: Just cleaned. Not worried about my house credentials around many repositories, I change them from time to time ;)

examples/www-image/main/jpg-render.c
```diff
@@ -0,0 +1,514 @@
...
+EpdiyHighlevelState hl;
+
+// WiFi configuration
+#define ESP_WIFI_SSID     "WLAN-724300"
+#define ESP_WIFI_PASSWORD "50238634630558382093"
+
+
+// Image URL and jpg settings. Make sure to update WIDTH/HEIGHT if using loremflickr
+// Note: Only HTTP protocol supported (SSL secure URLs not supported yet)
+//#define IMG_URL "http://loremflickr.com/1200/825"
```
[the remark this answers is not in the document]
martinberlin: Update: Works with IDF 4.2
I will make a post in esp32 asking why this way without certificate hangs and resets in 4.4 (latest master)

vroland:
Regarding streaming decode: Looks like the maximal size of a chunk read for decoding is defined in https://github.com/espressif/esp-idf/blob/166c30e7b2ed1dcaae56179329540a862915208a/examples/peripherals/lcd/tjpgd/components/tjpgd/include/tjpgd.h
as JD_SZBUF. So our buffer doesn't really need to be smaller than that.
Then, you could start the decompression before the post hand have a counter that tracks how many bytes are in the buffer. This counter and the buffer must be guarded by a mutex, so the decompression function cannot read while the http handler is writing the buffer.
If new data is comes in, copy it to the decompression buffer after the last still remaining byte. If no more data fits, sleep for a few ms in the http handler to let the decompression do its work, then continue filling the buffer if it is now free.
On the decompression function side, read the requested number of bytes from the buffer, decrease the counter and memmove the remaining bytes to the start of the buffer. If not enough bytes are available, sleep and wait for the http handler to fill the buffer, then try again.

vroland:
Regarding the bitmap example: Maybe we can find a way to re-use the http and wifi code and move the decompression / decoding functions to a different file? We could even detect the file type through the file's magic start bytes and support both.

martinberlin:
Is a good idea, but in my opinion will look messy, since the BMP does kind of progressive download + buffering the pixels to EPDiy directly. 
I vote for the complete deletion of the BMP example, unless you really want to keep it, then we leave it as a second isolated example. I would prefer to use this time to make the JPG decompression even faster, or try this that you commented before about decompressing while downloading, if you agree of course.
If we ever do something like detect file type, I would prefer to add PNG, that at least is a more modern image type and I'm sure more developers will be interested on having it available.

examples/www-image/main/jpg-render.c
martinberlin: Added, thanks!

examples/www-image/main/jpg-render.c
martinberlin: Added thanks!

examples/www-image/main/bmp-render.c
martinberlin: I forgot to add it. But is interesting to see that refreshes and works even without calling this.

martinberlin:
A small but relevant improvement: Remove [delays in feed_buffer (not needed) and the output_buffer](https://github.com/martinberlin/epdiy-rotation/commit/71a46596f8b13d57125bdc5f8cf8f2242d05dff3) 

48772 bytes read from https://loremflickr.com/960/540

I (10676) decode: 757 ms . image decompression Improved!
I (11401) render: 297 ms - jpeg draw
I (11402) www-dw: 1728 ms - download -> slightly faster
I (12621) total: 2782 ms - total time spent

vroland:
I agree, I would leave out the BMP version for now and rather polish the JPG version :)

martinberlin:
What is still missing from my TODO: 
Check if it’s possible to decompress src_buffer while downloading

>JD_SZBUF size: 512 bytes. So our buffer doesn't really need to be much bigger than that.
Then, you could start the decompression before the post hand have a counter that tracks how many bytes are in the buffer. This counter and the buffer must be guarded by a mutex, so the decompression function cannot read while the http handler is writing the buffer.

Will start to experiment on how to do this. I imagine that is alike the tasks as feed_display in EPDIy that is always looping with a semaphore. And this task will be in charge of JPG decompression, need to give it a couple of thoughts. If you think is a good idea, will come back to my CMakeLists
```
# Select only one app_sources
set(app_sources "jpg-render-advanced.c")
#set(app_sources "jpg-render.c") 
```
So we can keep a simple version and an advanced where we experiment and try to push it to the limits. jpeg-render.c as it is right now, with SSL working and all the options quite easy could be already used.
Will start with the advanced as soon as my other job that is not C or electronics, gives me some time to think.

vroland (approved):

mcer12:
@vroland  @martinberlin 
To make https work without checking the certificate, I had to go to:
`menuconfig -> Component config -> ESP-TLS`
and check the following:
`* Allow potentially insecure options`
`* Skip verification by default`

Maybe would be worth adding to the sketch / documentation? I will also try to recreate the sketch in arduino api so that users without knowledge of esp-idf can implement it as well.

martinberlin:
Hello @mcer12 
For me it was not needed to do that just leaving the define:

    VALIDATE_SSL_CERTIFICATE false

And this in turn does not add the .cert_pem property in the http_config struct. This for me was enough to get it working without any special menuconfig setting. But if it's required should be added, I would like to know only why in my case was not required, but for you didn't work out.

mcer12:
AFAIK ESP-IDF 4.2 and newer, you need to explicitly set it in menuconfig, otherwise it throws an error. What version do you use?

martinberlin:
I use 4.2 and for me it was not necessary to add this. Simply getting out the cert_pem from the http_config was enough. I don't know if @vroland had to add this additionally but I think not, our conversation is on the code channel of Slack.

mcer12:
@martinberlin Yeah it was 4.3+, my mistake :) I think the arduino examples should be made separate from the idf ones instead of the current workaround. As of now I only use esp-idf for epdiy, I don't have the need for it otherwise for my simplistic projects :)

PR #104, opened 2021-08-28
Fixed connector code

mcer12:
Wrong connector was listed - male instead of female, I found out just by ordering a wrong part... 
AXT434124 - male, on the display flex cable
AXT334124 - female, on the board, that's what we want


vroland:
Argh, stupid mistake. Thanks for the correction!

Issue #117, opened 2021-10-04
Invalid UTF-8 sequences cause a crash in utf8_len


cgreening:
I think the check in the for loop for seeing if you've reached the end of the UTF8 table is not quite right:

https://github.com/vroland/epdiy/blob/master/src/epd_driver/font.c line: 46

```
  for (utf_t **u = utf; *u; ++u)
```

This will run off the end as the `*u`  is not zero. It should be something like:

```
  for (utf_t **u = utf; (*u)->mask; ++u)
```

Which will run through the entries until it hits the last entry which has the structure all set to 0.

vroland:
Good catch, thanks :)

PR #118, opened 2021-10-04
Fix crash on malformed utf8

cgreening:
Fixes https://github.com/vroland/epdiy/issues/117


martinberlin:
Chris if I see this correctly this is already added in this pull request:
https://github.com/vroland/epdiy/pull/113/files#diff-4475865b68b6904e77f49c4140284dec978eb3af832ee432298e02b1474e0269

PR #129, opened 2021-11-30
LilyGo T5 4.7" Waveforms, recently open-sourced

entozoon:
LilyGo have kindly [open sourced](https://github.com/Xinyuan-LilyGO/LilyGo-EPD47/issues/32#issuecomment-948491023) their their [waveform](https://github.com/Xinyuan-LilyGO/LilyGo-EPD47/blob/master/Waveform_header/ED047TC2.h), so under their license we can go ahead and include it!

Thanks for the incredible work you've put in here @vroland , I hope I've used your variable style correctly. It sure works great; nice and crisp!


vroland:
Sure, I didn't get around to adding it yet, so thanks for the PR!

PR #131, opened 2021-12-10
Adding library.json for Platformio ESP-IDF build & new JPEGDEC example

martinberlin:
Simply adding this JSON so it's easier to use this component without having to make a subdirectory and including it directly.
@vroland I added also in this example a new version of JPEGDEC that is faster and it's the one that uses Larry Bank decoder:
https://github.com/bitbank2/JPEGDEC

I could add this component also here, but I'm not sure if the licenses are compatible, since it uses the [Apache License](https://github.com/bitbank2/JPEGDEC/blob/master/LICENSE). 
Please make a review and tell me if that is in order.

4F050087-3AB7-44B7-93AD-09D91CAD6E61

Performance test: 
```
60747 bytes read from https://loremflickr.com/960/540
I (12425) decode: 291 ms - 960x540 image MCUs:34
I (12425) www-dw: 322 ms - download
I (12425) render: 170 ms - render
I (14006) total: 783 ms - total time spent
```

Experimental option with JPEG_CPY_FRAMEBUFFER true
Note: It copies directly the whole row of decoded JPG into the framebuffer, hence looses Rotation and the gamma correction. The only benefit is that is ultra fast, decoding same image from exampe above:

I (12889) render: 1 ms - render so it will use 169 ms less to render the same image.

Update: after checking out the library I see that the actual work is done in C so maybe this can be refactored to use only C. But first I would like that someone tries it out and shares the experience with us.


vroland:
Hi, that looks interesting, thanks :)
From a quick look, the license should be fine: https://www.apache.org/licenses/GPL-compatibility.html
Though I don't have time to review the code in detail right now. I hope it's fine if I come back to this in the new year, maybe someone else wants to give some feedback until then :)

martinberlin:
@vroland it will be ok if here I use a **git submodule**
so I don't have to copy Bitbank2 files and instead I use his library the IDF way as a submodule?

Just made a [PR in his repository](https://github.com/bitbank2/JPEGDEC/pull/32/files) so we can make that but I wanted to know if that is ok.

vroland:
Yes, a submodule sounds good. Also that performance difference is definitely significant. Thanks!

martinberlin:
Great thanks. Is already added as a submodule. Only missing that someone gets to test it. Maybe @mcer12 once this busy end of the year goes away and 2022 starts :)
D362F0C1-7263-4492-9772-A477E2AD1969

vroland:
Hey, I tried to get it running today, but I ran into some problems:
First, my compiler (IDF 4.4) won't compile the jpgdec version because it's c++ and the compiler complains about field
order in line 317. I can fix that, but then cerificate verification does not work, wile it works in the jpg-render.c version.
Ignoring that, it works :)
Did you change anything about the connection between the versions?
That brings me to code duplication: I think it would be nice if we had one source file and depending on which library you
want to use you just include a different header. E.g. make a jpegdec and tjpgd decoder header and expose
their functions with extern "C". Then have just one main c file that can use either one or the other.

It's definitely nice to see the faster decoding! I just think the example is a bit complicated to set up right now. Do you think you can remove the duplicate code?

mcer12:
@vroland @martinberlin I don't think it's a good idea to make it an idf component because it's not easily applicable in arduino. I know you guys use esp-idf environment but people like me prefer the arduino environment where esp-idf components are fixed. As a matter of fact, I use arduino environment in platformio with epdiy and it works perfectly well! No need to use arduino as idf component. Why not just use the header file and use it like any other lib?

martinberlin:
Thanks for the review:
>First, my compiler (IDF 4.4) won't compile the jpgdec version because it's c++ and the compiler complains about field
order in line 317.
In Line 317 from what file? I'm using so far 4.2 to build but I can update it and try again. 

>E.g. make a jpegdec and tjpgd decoder header and expose their functions with extern "C". Then have just one main c file that can use either one or the other.

Sure we could add some kind of wrapper around JPEGDEC that uses only the C part instead of adding the CPP and then both stay in this language without even using C++. The C plus plus part is only a wrapper if I see that right.
If that's done then at that point you could switch header and it will implement one or the other decoder.

> @mcer12 said: I don't think it's a good idea to make it an idf component because it's not easily applicable in arduino.

I don't get this point. If Platformio finds the files it should be able to compile this as an external library just as it does with EPDiy or any other library. What part of adding a component is not applicable in Arduino?
Can you please elaborate a bit more your statement so I understand better what you meant?

martinberlin:
@mcer12 after my last commits you should be able to test this since it's not working anymore as a submodule. 
@vroland I was starting with a preliminar analysis to see if we can gain anything making a header to switch between encoders:
Research about JPEG header to use JPEGDEC as C

Used functions:
jpeg.openRAM(source_buf, img_buf_pos, JPEGDraw4Bits)
jpeg.setPixelType(FOUR_BIT_DITHERED)
jpeg.decodeDither(dither_space, 0)
jpeg.getLastError()
jpeg.close()
Dithered is done by the decoder using decodeDither

VS TJPGD
jd_prepare(&jd, feed_buffer, tjpgd_work, sizeof(tjpgd_work), &source_buf)
jd_decomp(&jd, tjd_output, 0)
Dithering has to be done outside TJPGD

And they are kind of very different functions with parameters that are not exactly interchangeable, but also with the difference that the dither for example, is already included in JPEGDEC but is not in TJPGD (Is done externally)
So I'm not really sure if we can gain anything trying to avoid "code duplication" and instead end up with a hard to understand single C file, with a lot of preprocessor directives. Anyways as you mentioned, it deserves a try, maybe with the help of more experienced colleagues ;)

vroland:
Sorry, I kind of let this slip... I'll have a closer look until the end of the weekend!

vroland:
True, the library interfaces are different, but we can de-duplicate the defines, wifi and HTTP code. 
So basically, we could move the `decodeJpeg` resp.`drawBufJpeg` and dependencies to their own files. 
Then we could have a main.c, tjpegd_impl.h/c and jpegdec_impl.h/c? Buf if that's too complicated I would at least suggest moving some defines to a common header, like settings.h, for the common settings. 

Ohterwise, nice work so far! I think the defines are clearer now and it works for my board if I enable the "use spiram in malloc()" option in the IDF. I suspect that with the larger display, there is not enough memory for TLS otherwise.

There is one strange thing I noticed though: The time to refresh the display (epd_hl_update_screen) takes twice as long in the JPEGDEC example than in the TJPEGD one. I this code should be the same for both, except one is called from C++? But then it makes no sense for it to be slower. Can you reproduce this?

martinberlin:
Hello Valentin,
Thanks for your comments. I moved the common configuration to **settings.h** and it looks cleaner now. Also enables for a fast switch between examples. 

I like that both examples now are kind of self-contained and I think we won't gain too much adding separated dependencies for both (Although we could do it there is more to come) But I like the simplicity how it works at the moment. 

>There is one strange thing I noticed though: The time to refresh the display (epd_hl_update_screen) takes twice as long in the JPEGDEC example than in the TJPEGD one. 

I found an unnecessary delay, can you pull last changes and try again?

About certificate validation in ESP-IDF version >= 4.3 the setting VALIDATE_SSL_CERTIFICATE needs to be on true "as default" otherwise will give a configuration error:

```
E (8030) esp-tls-mbedtls: No server verification option set in esp_tls_cfg_t structure. Check esp_tls API reference
E (8030) esp-tls-mbedtls: Failed to set client configurations
E (8037) esp-tls: create_ssl_handle failed
```

I think they assume after this version that SSL means validation of the certificate. For those who want to skip that they can simply use HTTP only or also I added in the Readme section information on how to avoid this validation in menuconfig. 
That said, I think you can give it a last try and see if there are further corrections, both should go to sleep after render with the mission to consume at least as possible.

vroland:
Thank you, I still wonder why the screen update is slower with JPEGDEC, but maybe that's something to investigate for another time.

Issue #148, opened 2022-02-02
Make this a PlatformIO Library


matteocarnelos:
It would be awesome for this library to be a PlatformIO Library, available in the PlatformIO Registry.

It should be enough to properly fill the `library.json` file as described in [Creating Library - PlatformIO](https://docs.platformio.org/en/latest/librarymanager/creating.html).

vroland:
Right, that would probably make sense. It would need some testing though. And it would be nice if it could be used seamlessly
with either the arduino framework or esp-idf through platformio. I don't have the time to work that all out right now, but I'm happy to test if someone creates a suitable metadata file.

martinberlin:
@matteocarnelos after the merge is there anything more to be tested here or we can close this issue?

matteocarnelos:
I think everything's done and this issue has been solved

PR #151, opened 2022-02-07
Update library.json for PlatformIO Registry

matteocarnelos:
Hi! I added to the `library.json` all the necessary fields to publish this library on the PlatformIO Registry (#148).
A lot of info have been taken from the existing `library.properties` file.

I tested it and it seems to work just fine to me.
If you want to publish it you can register a new PlatformIO Account and then use the `pio package publish` command. 

#### Extra:
I also added a Github Workflow to automatically publish the library to the PlatformIO Registry at every new tag on the repo.
If you want to use it you just have to set the `PLATFORMIO_AUTH_TOKEN` secret on the repo. Otherwise you can just delete the workflow.

#### Note:
Remember to update the version of the library in the `library.json` file at every publish, otherwise you'll get an error.


matteocarnelos:
Sure! I'll apply the suggested changes.

vroland:
Thanks!

vroland:
The package should be published now, so feel free to test :)

schroedi83:
Hi! I tested it and it worked for me. The only unusual thing is that for example the img_beach resource is shown as if it would be an example itself.
Bildschirmfoto vom 2022-02-19 15-24-42

matteocarnelos:
You're right @schroedi83. I think we should specify the examples folder in `library.json`.

Not sure if this is enough:
```json
"examples": [ "examples" ]
```
Or if we have to specify every example:
```json
"examples": [
    {
        "name": "calibration_helper",
        "base": "examples/calibration_helper",
        "files": ...
    },
    ...
]
```

matteocarnelos:
Ok I made some tests and it seems to be a problem of the PlatformIO examples visualizer, as it displays them without keeping the folder structure. Instead, the web version of the PlatformIO Registry displays the examples correctly (you can check it [here](https://registry.platformio.org/libraries/vroland/epdiy/examples)).
Therefore there's not that much that we can do. Maybe in the future the PlatformIO devs will switch to the (better) PlatformIO Registry web visualization.

Issue #136, opened 2022-01-09
Lilygo T5-4.7 - Current leaking to 3v3 rail after epd_poweroff


kaweksl:
Hi,

On Lilygo T5-4.7 powered with battery after screen init and epd_poweron() board takes ~115mA@3.6V , and after epd_poweroff() but with ESP32 fully running ( NOT in deep sleep ) power consumption increases to ~235mA . 

Looks like some control pins on ESP32 are staying high after epd_poweroff(); and some current passes from esp32 via epd chip clamping diodes to switched off 3v3 rail . This is not happening before epd initialization .

I have noticed after epd_poweroff() esp32 pins IO26 (STH) , IO5(CKH) and IO2(EP_D4) have ~2.2V what may indicate high current draw and drive overload .

Current draw measured with PPK2

EDIT: I need to clarify Lilygo T5-4.7 have rail called VDD3V3 that is powering ESP32 and shift register , it also have 3V3 rail that provide power to epd and LT1945, that rail is controlled by shift register and epd_poweron/poweroff . My theory is that when epd is off, current from esp32 pins go through clamping diodes of epd chip to not powered 3v3 rail and from there wants to power up LT1945 . 
This doesn't happen if powered by USB because then EPD and LT1945 is always powered on .

mcer12:
First off, this is not the place to debug lilygo board. Even though it's supported by the library, you should open an issue on lilygo page, not here.

It doesn't make any sense for the board to draw double the current when the power to the display is off compared to when display is on. The pins can't even draw such current because to my knowledge ESP pins are limited to around 20mA max. Also, these are signal pins and there's no way the display would draw such excessive current through data lines.

The Lilygo board is not particularly great, it's based off of old outdated epdiy design and it has issues with the display fading out when sleeping.

kaweksl:
As far i know ESP32 has Cumulative IO output current  = 1200mA and max ~40mA per IO, that also depends on pin power domain. I have tracked down 3 pins, 3x40mA = 120mA..
What if VDD = 0V, screen (driver IC ) data lines have clamping diodes to VDD (common protection method) , voltage on data lines is higher than on VDD, current is passed trough diode to VDD, something else on VDD is trying to draw that current as not only screen is connected to VDD .

![Screenshot from 2022-01-19 22-54-56](https://user-images.githubusercontent.com/8569165/150220667-d439a86f-65d7-46a8-834c-9fa7d23dd0f4.png)



[...]

PR #154, opened 2022-02-10
Added i2s gpio attach/detach on epd_poweron/off

kaweksl:
bug fix for current leaking issue: https://github.com/vroland/epdiy/issues/136#issuecomment-1035410191

I do some changes in i2s_data_bus* :
- removed use of "i2s_bus_config i2s_config" as only variable it carry is epd_row_width, rest are globals anyway,
- added i2s_gpio_attach/detach

In display_ops.c added calls to i2s_gpio_attach/detach on epd_poweron/off, but only for Lilygo T5 4.7 board. 

Question is if this should be limited to Lilygo as this shouldn't harm other boards.

### DIFF
--- a/src/epd_driver/display_ops.c
+++ b/src/epd_driver/display_ops.c
@@ -95,28 +95,25 @@ void epd_base_init(uint32_t epd_row_width) {
   push_cfg(&config_reg);
 
   // Setup I2S
-  i2s_bus_config i2s_config;
   // add an offset off dummy bytes to allow for enough timing headroom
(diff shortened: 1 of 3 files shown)


vroland:
Hi @kaweksl, thanks for the PR! 
I think I'd attach / detach for the other displays as well. As you said it won't hurt and seems conceptually cleaner to me.

vroland:
Just confirmed with a V6 board, seems to be working nicely! :+1:

martinberlin:
@kaweksl is there anything else to be checked on this issue after the merge, or it can be closed?

PR #155, opened 2022-02-12
boards.local.txt flash fix for the LILYGO 4.7" EPD

Hunter1753:
This MR fixes an issue for the LILYGO 4.7" EPD where only 4MB partitioning schemes could be uploaded successful, because the flash size is hardcoded to 4MB.

The LILYGO 4.7" EPD module only comes with 16MB.


vroland:
Good to know, thanks!

PR #157, opened 2022-02-17
Handle larger fonts

thigger:
Unfortunately #147 wasn't quite enough for me (font size 200 for a clock display) - there were a few more 8-bit integers that needed changing to get things displaying properly.

Related to #94 for really big fonts.

### DIFF
--- a/src/epd_driver/font.c
+++ b/src/epd_driver/font.c
@@ -126,7 +126,7 @@ static enum EpdDrawError IRAM_ATTR draw_char(const EpdFont *font, uint8_t *buffe
   }
 
   uint32_t offset = glyph->data_offset;
-  uint8_t width = glyph->width, height = glyph->height;
+  uint16_t width = glyph->width, height = glyph->height;
   int left = glyph->left;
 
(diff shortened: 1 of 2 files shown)


vroland:
Maybe you should also increase the size of the compressed_size field then?
I hope this doesn't increase the memory footprint too much, though the original struct was not packed anyway.

thigger:
Happy to do so - for my own use-case this is sufficient, but I can see that it's not far off being a problem - in the uncompressed version of the font I'm using (which won't actually fit on the controller) the largest value of compressed_size is 47318 which suggests this would be an issue somewhere around font size 230 with uncompressed fonts.

thigger:
I've made the change, but it strikes me that compressed_size is only used for compressed fonts, which are unlikely to break 65535 bytes per glyph - and that the issue could equally be solved by making fontconvert.py set compressed_size to 0 for uncompressed fonts (and leaving compressed_size as uint16_t)

vroland:
That's true. Ok, then maybe we should leave it at 16 bits. I'm not sure if drawing huge fonts is a common use-case anyway.

Issue #92, opened 2021-07-10
Lilygo T5-4.7 screen turns darker grey while on battery


rmk92:
I've just connected a battery to my Lilygo T5, and I notice that as soon as the power is cut (blue LED turns off) the white areas of the display starts turning darker. I'm using epd_power_on() and epd_power_off() to just power the display during updates. This doesn't happen when connected to USB, when the hardware ensures that the display is cosntantly powered.

I haven't investigated with a 'scope yet, but I suspect the issue may be that the e-ink display isn't being allowed to properly power down. Does epdiy guarantee that the signals to the e-ink display get set to logic low or outputs disabled while the power to the e-ink display is turned off?

There are other issues raised with this which suggest people eventually see their display failing - this might be explained if the display is remaining semi-powered through the logic signals? Please remember that it is quite normal to have diode clamps to the power rails on silicon devices.

vroland:
Interesting theory, I didn't really consider that it until now. I mainly used the e-reader screens and I never had an issue with them.
Could you try calling `epd_deinit()` after turning off the power to the display? Maybe you need to additionally set the data pins as inputs to isolate them after that. If that changes anything, we should probably add that to the driver.

mcer12:
@rmk92 @vroland  powering the display with signals is not the case, those go to the display controller which then controls mosfets with high voltages to change the pixels. You can just cut the high voltage rails and see what the display does without them :D 

What I suspect is that when you just cut power, the display power rails are not shut down in correct order and that's why you see this phenomenon. Bottom line is - for the display to work correctly, you need to use deep sleep. Your other options would be to sense power outage, and immediately run the display shutdown and deinit functions (you probably won't have time for that before the capacitors drain). Or simply shutdown & deinit before you disconnect power - with a button or whatever.

lixy123:
as  the code of "https://github.com/vroland/epdiy/tree/master/examples/lilygo-t5-47-epd-platformio",
if you use battery power and not usb power. When the ink screen enters the deep_sleep after call "esp_deep_sleep_start", a large number of black spots will be added at the white display of the display screen, as shown in the following figure

before esp_deep_sleep_start:
![2-81](https://user-images.githubusercontent.com/35212560/152978045-04d253b2-8f4a-4abd-a893-f4a9fb9c911a.jpg)

after esp_deep_sleep_start:
[...]

PR #158, opened 2022-02-18
Lilygo config register should be set to low on poweroff()

martinberlin:
Simple update only for V2

Lilygo config register should be set to low on poweroff() 

Related issue #92


kaweksl:
Much better. I had issue with screen background greying after epd_poweron/off cycle without any pixel change on screen. Now it seems fixed

kaweksl:
Much better. I had issue with screen background greying after epd_poweron/off cycle without any pixel change on screen. Now it seems fixed

EDIT: or maybe not fixed, gonna have to test it more

Issue #160, opened 2022-02-21
Move away from defines


mcer12:
In particular board revision, display type and display resolution.

I think using defines / esp-idf settings wasn't the best way to make the library portable. In arduino IDE it's a huge headache to make it work. In platformio it's fine but still little overly complicated for a layman. What I see as a better approach would be selecting the display type and board version in epd_init. This would work on all platforms, there would be no need to go in menuconfig or add custom boards in arduino ide. Similar to how most other libraries work like GxEPD or u8g2

Proposed solution would look like this:
**epd_init(EpdLutOptions lutOptions, EpdDisplayOptions displayOptions, EpdBoardOptions boardOptions);**

So something like this:
**epd_init(EPD_LUT_64K, ED097OC4, EPDIY_V5);**

I am also thinking about adding an override for display resolution so if you have an unusual display, you can use this:
**epd_init(EpdLutOptions lutOptions, EpdDisplayOptions displayOptions, int displayWidth, int displayHeight);**
instead of digging in the library.

What do you think @vroland @martinberlin

martinberlin:
Hi @mcer12, in my humble opinion if this could be an alternative I think it could work fine, not sure if it's possible to do it alternatively while maintening the existing way. Below my thought's about it.
The problem I see if the **epd_init** is updated so it works only in this new way is that it breaks backwards compatibility (BC) with all what is already built on top of this library, including existing examples.
The **idf.py menuconfig** has also the benefit that some options can turn different menu's on and off, like currently happens with V6 board (VCOM option appears only when this board is selected)
And also, I think, Kconfig offers an easy way to see what defines are being set. Since I see all people that implemented this component using Platformio correctly set the Board and Display options.
@vroland It would be possible to mantain the current options to avoid breaking BC with all the rest, while offering alternatively an option to do an custom epd_init ?

That way maybe it could be easier in the cases mcer mentions while not breaking existing working Firmware's that are build with EPDiy. I've been always a bit scared of making updates if they break BC because it can potentially break existing Firmware (Thinking for ex. about Platformio dependencies, where EPDiy is "picked up" and installed in the lib_deps folder automatically)

mcer12:
Yes, this is a breaking change but changing the sketch takes below a minute no matter how complicated, it's a minor annoyance really. We can always add a custom compiler error with a notice that there's been a breaking change with a link to wiki. I think in the end, it would simplify library usage for the user. Epdiy is just difficult to use at this point, especially with arduino ide.

[...]

PR #166, opened 2022-03-01
Refactor/160 EpdDisplay part

martinberlin:
Hello,
This is my proposal to merge the dynamic display configuration. After this you don't need to use menuconfig for the display. You just insert the display like you can see in the Dragon example. 

This PR is just a declaration of intentions that partially addresses #160 . The dynamic board config part will be proposed by **Micke Prag** (In epdiy.slack.com)

It should be extensively tested and if approved then I will go and update all examples since now every example should inject at least the proper EPD you are using for the demos.


martinberlin:
At this point I'm not sure if this is the best way to go. Will consult it with the oracle (my pillow) for a few weeks and then we can move forward.

martinberlin:
Hello @vroland @jdoubleu @mickeprag 
Shall we ditch this and forget about it?
Concept was interesting but I’m not sure if it makes things easier also for stale after being hanging here for so long. Feedback?

jdoubleu:
So these changes would allow me to select (and switch) the display at runtime rather than at compile-time (through ifdefs)?

I don't see a real benefit here. I don't have a use case, where switching the display type midway, applies. I'm not even sure if that's even possible (maybe if a display type somehow supports multiple different configurations and waveforms).

The only other feature I could imagine is having more than one display connected to the ESP. Then it would be very hard to select which ones to include at compiler time. However, neither the board, nor the driver support that. And I don't think anyone would need that feature anyways.

Furthermore it increases code size, although slightly, because all those definitions (and waveforms) have to be available at runtime, so they must be included in the binary.

Is there any possible use-case I missed? Otherwise I wouldn't introduce this feature, as long as there is no real benefit in doing so.

martinberlin:
Well basically all this was because in #160 "Move away from defines" I though that it could be interesting to also declare the display this way.
Let's close it with this comment as there are many other much more important to topics than this and as said previously is already too long here. Thanks for your feedback.

mickeprag:
One use case for me is that I have one firmware for multiple boards. So the firmware can detect at runtime what kind of display is connected and initiate the correct one.

martinberlin:
This branch will remain on my fork. So if one day we want to implement it we can do it using the latest master

PR #167, opened 2022-03-02
added a function epd_get_string_rect into fonts.c

Yardie-:
I have added a function epd_get_string_rect into fonts.c 
This allows works with multiple lines and also has a margin parameter to allow simple text area margins.
Unfortunately my rough hack to handle the touch function on the lilygo is also in there.
Both do actually work as expected.
Thanks Greg

### DIFF
--- a/src/epd_driver/config_reg_v2.h
+++ b/src/epd_driver/config_reg_v2.h
@@ -68,7 +68,6 @@ static void cfg_poweron(epd_config_register_t *cfg) {
 
 static void cfg_poweroff(epd_config_register_t *cfg) {
 #if defined(CONFIG_EPD_BOARD_REVISION_LILYGO_T5_47)
-if defined(CONFIG_EPD_BOARD_REVISION_LILYGO_T5_47)
   // This was re-purposed as power enable.
   cfg->ep_scan_direction = true;
   // POWEROFF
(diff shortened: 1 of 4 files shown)


vroland:
Looks good to me, thanks :) Unfortunately I can't test this right now, so hopefully someone complains if it causes problems.

ashald:
This prevents Lilygo from going to deep sleep properly and its pwoer draw remains at 50mAh.

epdiy_lilygo_power

It looks like there is `epd_pwoeroff_all()` that supposed to be called instead? Bit it's only available when Lilygo board definition is set, which means that library users not only have to set the config definitions, but also have ifdef in their code. As opposed to if `epd_pwoeroff_all()` would've called `epd_poweroff()` by default, and would've done some special logic for Lilygo.

Yardie-:
The problem was that without the power_off and power_off_all there was no way to maintain the touch when the epd power was turned off. I will look into it and see if I can find a better solution. It was an unintended addition. It works in Arduino as expected however clearly not in ESP_IDF

mickeprag:
This change also affect my PR (#160). Could the code in `power_off_all()` always be called instead of `power_off()`. So there is only one power off function?

martinberlin:
@mickeprag Feel free to do any required change so we merge #160 then we can add this later. 
Now this function is called:  [epd_powerdown()](https://github.com/vroland/epdiy/commit/3ffb83c33cf424494c07fcd2715ba9123ffc64d3) and it's actually only used for Lilygo board.

Actually this is the only board that is bad designed and needs an additional function to power down everything, except the touch, so you can wake up with that...

mickeprag:
I am sorry, I saw it was removed after my comment.

I think I found a solution/workaround in my PR #160. See my lasts comment. Please disregard this question here.

PR #168, opened 2022-03-07
160 remove board defines

mickeprag:
This is my take on #160 to remove defines for board definitions.

The goal has been to not break backwards compatibility. All current project should still work without modifications. Configuring the board to use is still possible through `menuconfig`, but is not required. For projects not using this library as an esp-idf component there is a new option (and default if nothing is defined) called  `EPD_BOARD_CUSTOM` that can be used by PlatformIO projects or Arduino.
These project must call the new function `epd_set_board()` before `epd_init()` to tell which board to use.

The different code for each board is now separated into own source files to keep maintenance as clean and easy as possible. For third party boards these can also be used as a templates since it will be possible to support new boards without modify epdiy source.

Since this branch touches quite many files I have tried my best to create isolated atomic commits for easier review.

This is currently only tested on v5 boards.


martinberlin:
Tested with Lilygo EPD47, just pulled Micke's repository and switched to branch **160-board_hal**
Then I just run menuconfig and selected right board and display. Build reports this: 

```
/home/martin/.espressif/tools/xtensa-esp32-elf/esp-2020r3-8.4.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: esp-idf/epd_driver/libepd_driver.a(render.c.obj):(.literal.epd_init+0x4): undefined reference to `epd_board'
```

mickeprag:
Sorry, my mistake. Please try again now.

martinberlin:
Thanks Micke, now is complaining about this:

**/src/epd_driver/board/epd_board_v5.c**:5:10: fatal error: display_ops.h: No such file or directory
 #include "display_ops.h"
          ^~~~~~~~~~~~~~~

Tomorrow when I find some time I will go through your code and try to get more awareness of what you modified. Now if you go into the referenced files, for example in my board **epd_board_lilygo_t5_47.c** and I change the includes to reference this folder:

```c
#include "../display_ops.h"
#include "../i2s_data_bus.h"
#include "../rmt_pulse.h"
```

Then it compiles and by flashing I get this error:

assertion "epd_board != NULL" failed: file "/home/martin/esp/projects/pull-requests/epdiy-micke/src/epd_driver/render.c", line 249, function: epd_init

mickeprag:
Fixed these as well. Thanks.

martinberlin:
But did you push the changes? Or it’s me that I don’t find the commits

mickeprag:
I did push, yes. But there is no new commits.

martinberlin:
I ask because I don't see the changes pulling from origin, same error as before is reported:
epd_board_common.c:1:10: fatal error: ../epd_board.h: No such file or directory
 #include "../epd_board.h"
          ^~~~~~~~~~~~~~~~
          
Sorry but changes are not pushed, last commit in the log:
commit d7c2ced9d35da2a61430609d8f3ac305de83f5d4 (HEAD -> 160-board_hal, origin/160-board_hal)
Date:   Mon Mar 7 21:41:45 2022 +0100

mickeprag:
You have the change. The path did not contain ".." before. Why you still get compile error I need to check though.

The last commit has not been changed so the date is correct.

martinberlin:
No problem. But I must say I really don't get how is updated. Because usually if you review something, comes a new commit + push and you can check it before building it.

martinberlin:
Started to play around with this so I can use it for the next revision of my S2 board
[epd_board_v6_s2.zip](https://github.com/vroland/epdiy/files/8274648/epd_board_v6_s2.zip)

so you can check it out. I might have forgotten something.

Apart of that, in order to use this MCU, this are the things that need to be updated and I've on my list:

- i2s_data_bus.c  S2 has only I2S0 (only one I2S peripheral)
- epd_board_common.c:13:33: error: 'ADC_WIDTH_BIT_12' undeclared (first use in this function); did you mean 'ADC_WIDTH_BIT_13'?
- rmt_pulse.c:42:18: error: 'ETS_RMT_INTR_SOURCE' undeclared

vroland:
Thank you for going forward with this and sorry for the late reply. I like the direction this is going :)
Currently it looks like setting the VCOM voltage for V6 still only works trough defines. We shoudl probably expose an "epd_set_vcom" function in epd_board_specific.h?

mickeprag:
Thank you @vroland. I have some ideas regarding the VCOM voltage. But previously I had no way to test it since I don't have a v6 board. I just got my own version running so I should try to get some solution up on Github shortly.

vroland:
Sounds good :)

vroland:
Since I'm currently travelling I can't test this right now, so I'll have to rely on the judgement of the other contributors to try this out..

martinberlin:
I will be also kind of busy at least until end of April, beginning of Mai since I'm reallocating to Barcelona and have a big mess in front of me. 
But this should be tested at least in:

- [ ] V6 boards (@mickeprag  has one clone)
- [ ] V5 boards @mcer12 
- [ ] LilyGO this I could test

Let's keep it active and test it good before doing any merge. Also IMHO this should get a v1.0 tagged release and after this update it should have a different release version, so people using Platformio can relate to this using the: #v1.0 tag at the end of the github URL to avoid breaking any existing build. Maybe it's not needed but I will take some precautions provided that this is kind of a big update and touches core files.

mickeprag:
I just rebased this branch on top of master and fixed the following:
 - `epd_powerdown()` that was only defined for LilyGO boards has been renamed to `epd_powerdown_lilygo_t5_47()` and is always defined. The old method is left to provide backwards compatibility but has been marked deprecated
 - To set the vcom voltage for v6 boards now have 4 options:
 - 1) Set the value in menuconfig (as before, for esp-idf projects)
 - 2) Define `CONFIG_EPD_DRIVER_V6_VCOM` (as before)
 - 3) Include the symbol `int epd_driver_v6_vcom` (as before, for Arduino)
 - 4) Implement the function `uint16_t epd_board_vcom_v6()` and return the voltage.

I feel I am pretty much done with this PR and is ready to merge.

martinberlin:
Awesome Micke, I will take some time to test it on the following days, and we can merge it. 

Question: Do you think we should make a new release in case the Platformio users can still use the old way just fixing it to a release tag? (Like #6.2 )
Or we really do not have to do that? CC @mcer12

mickeprag:
> That epd_board.h is in the include path. So I think is save to add it without the / backslash. Can you please change it?

Done

Issue #171, opened 2022-04-05
Greyscale mapping doesn't work as expected


aceisace:
Hi @vroland 
I've been testing the epdiy-library for quite some time, initially using the forked code of the 4.7" lilygo Display. When switching to this library, it seems there is a bug during rendering. The greyscales are not shown as expected. Some appear to be missing while others seem duplicated.

As a test and to help you with debugging, I first testet the greyscale test of lilygo:
This is how it should look like:
![1](https://user-images.githubusercontent.com/29558518/161867816-68f693ef-58a1-4286-9771-e0ca656857fc.png)
And this is how it is rendered with the lilygo library:
![2](https://user-images.githubusercontent.com/29558518/161868128-5c04d11d-c05c-40f5-87ba-7a55afe58cbd.jpg)

Hence it _should_ look the same with the epdiy-lib. But this is the result with the epdiy-lib:
![68717F8C-3E78-488B-8C32-26C38ED75783](https://user-images.githubusercontent.com/29558518/161868377-1e594c32-02a1-4887-8587-8bf45b9323dd.jpg)

To rule out that this issue was specific to the 4.7" display, I then tried it with the 9.7" one, with the following result:
![4](https://user-images.githubusercontent.com/29558518/161868645-0a94d402-d05c-468b-a852-5adb125f2a3d.png)

Could you look into this please? Thanks in advance

Yardie-:
I'm a bit curious about this one. 
I have a couple of questions though. 
What was each screen before the image render. That is did they both have a clean white screen prior to drawing the image.
Did you try redrawing the image over itself.
I don't know a heap but sometimes strange things happen depending on the previous state of the screen.
Also did you try the different refresh types.
What I see here is a waveform issue rather than a code issue. Lillygo did GPL their waveform and it is in the epdiy release as it stands. So I'm curious as to what is going on here. Also maybe you could add the files necessary ie code and image so that others can look into this properly. Otherwise your request is pretty vague and hard to reproduce.
Just some ideas.

aceisace:
@Yardie- Hi and thanks for your comment.
Some additional info:
* All displays were cleared (white) before displaying these images
* Since the images are generated using maths, issues related to buffer-generation can be ruled out
* I have tested both waveforms, the ones provided by epdiy and original waveforms. Changing the waveform doesn't cause any changes. The lilygo lib works as expected. But epdiy doesn't seem to render it correctly
* Redrawing the image over itself does not cause any changes

[...]

PR #172, opened 2022-04-23
Fix Lilygo EPD47 waveform adding ED047TC1 as default #171

martinberlin:
@aceisace please try this branch and tell me if the Grayscale now works correctly for you


aceisace:
Yes, it does work! I posted an image of it in #171 . Regarding this PR, both waveforms-header files xxtc1 and xxtc2 use the same name `epd_wp_ED047TC1`. Is this supposed to be like this? Second question, how can I use the alternative waveform file (xxtc2)? Do I need to update `boards.txt` or are there additional steps? Thanks in advance

martinberlin:
We could update the variable name to ED047TC2. And add it also in KConfig and as a constant so you can inject it using build flags, does that sound?

vroland:
Note that the timings may be temperature dependent as well. Though if you have done your tests at room temperature it should be fine.

vroland (approved):

martinberlin:
Thanks @vroland 
About the idea to add both Waveform in KConfig, is a good one?
Or we should leave this as before and add the new one as an **optional** one that can be added?

I just wanted to make thinks as simple as possible. If we add the two we should at least mark them correctly. I suggest something like:

ED047TC1 Fast waveform (similar to Lilygo repository)
ED047TC2 Slower with darker black (7 times bigger in data RAM)

Or something so we can differentiate them. Otherwise I suggest just leaving ED047TC1 and ED047TC2 as extra waveform that needs a small hand modification (uncommenting include and commenting the other one)

aceisace:
I'm in favor of the build flags option as well since it's more space-efficient and more intuitive which waveform was used for the 4.7". Do let me know if I can help with testing 👍

martinberlin:
>I'm in favor of the build flags option as well since it's more space-efficient and more intuitive which waveform was used for the 4.7". Do let me know if I can help with testing

@aceisace please check out latest update. Now I added this in the Kconfig

menuconfig_epdiy

@vroland since this waveform is strangely 7 times bigger, I'm adding the include file only if it's selected, is that alright?  Or it does not really matter?
I just though that we could save some more space for your own program if it is included only when you select that Display.

aceisace:
@martinberlin Thanks for the fast implementation! I'm only using Arduino IDE, will I be able to test the changes there too? If yes, I'll test it soon 👍 
Good luck with moving!

PR #175, opened 2022-05-07
add an experimental VCOM calibration example for v6+ boards

vdp:
Hi,
because the panels from the Ali very often come without a VCOM label, I wanted to use the TPS65185's built-in calibration feature to estimate the optimal voltage. The code is not ideal on many levels:

* firstly it uses relatively low-level EPDiy functions/constants/etc that are not exposed in the interface headers, so I suspect @vroland may not be very happy about breaking the interface/implementation barrier in this way. On the other hand it shows how if one needs to do it(not claiming that this is the case here, just speaking in general), they can easily reach deep into the internals. IMO this is useful.

* also, the measurement procedure requires to "drive the panel with NULL waveform". And of course this being eInk-related what this means exactly is a great secret. I've found that the voltage estimate is highly dependent on the driving time. I only have a single panel(with label) and that's what I've used to tune the time. Not sure if that will work reliably for different boards/panels. If someone can test it using their hardware that would be great.

* and finally I'm still learning the EPDiy code so the implementation may be completely wrong.


martinberlin:
Looks like an interesting addition @vdp 
Sadly I have not at the moment a V6 board at hand to test it so someone else should. Just tell me on epdiy.slack.com on any channel and I assign it to someone that can test it and give feedback about it.

vdp:
No worries. As you know I've already sent a message on Slack, so if someone is interested they will try it. This is just something I wanted to try and see if it works. Feel free to close this PR if no one chimes in.

mickeprag:
I also want to try this! I just don't have the time at the moment...

vdp:
There is no hurry whatsoever, as far as I'm concerned. It would be great if you could test it @mickeprag. It would be especially interesting if you could try with both 9.7" and 6" displays for which the "correct" vcom voltage is known.

mickeprag:
I am going to test this. Unfortunately I can only test agains displays without a label with a VCom voltage. So I have not idea what the "correct" vcom voltage would be...

vdp:
I don't think this is useful for anyone(me included) in its present form.

mickeprag:
I tried but got compile error so I did not manage to run it sucessfully

Issue #156, opened 2022-02-14
Epdiy not compatible with ESP-IDF v5


jdoubleu:
I noticed, that this project is not compatible with the upcoming ESP-IDF version 5. To be fair, that version is still in _early_ development and looking at their [support periods](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/versions.html#support-periods) can be estimated to be released mid to end 2022.

The migration guide can be found here: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/migration-guides/index.html.

Getting the latest ESP-IDF environment and compiling the [demo](https://github.com/vroland/epdiy/tree/master/examples/demo) can be done with the [`espressif/idf:latest`](https://hub.docker.com/r/espressif/idf) docker image:
```sh
git clone --recursive https://github.com/vroland/epdiy.git
docker run --rm -it -v epdiy:/app espressif/idf:latest sh -c 'cd /app/examples/demo; idf.py build'
```

<details> 
  <summary>Build log</summary>

```
[877/898] Building C object esp-idf/epd_driver/CMakeFiles/__idf_epd_driver.dir/display_ops.c.obj
FAILED: esp-idf/epd_driver/CMakeFiles/__idf_epd_driver.dir/display_ops.c.obj 
[...]

PR #177, opened 2022-05-07
Configure GitHub CI

jdoubleu:
This PR introduces a workflow for the GitHub CI, which will build all example applications. Since all of them require the `epd_driver` component, the driver is implicitly built as well.

In order for all examples to build successfully, I needed to update some of them. Most of the time it was just changing a default value in `sdkconfig.defaults`, to e.g. use the custom provided partition table (`partitions.csv`).

The CI will run on every push to the `master` branch or for every PR which targets the `master` branch.

The motivation and benefit behind building all examples in the CI is to validate changes made still compile with different versions of ESP-IDF. This begs one important question: Which ESP-IDF version do you want to actively support? The latest stable release (i.e. `v4.4.1` at time of writing)?

If the CI is not yet running for this PR, head over to my fork, where I ran it multiple times (during development): https://github.com/jdoubleu/epdiy/actions/runs/2286199069

As you can see, all examples fail to build with the latest version of ESP-IDF, as explaine din #156. If you would merge this PR now, it would fail for every push. I suggest to create a branch off of this one, to fix the incompatibilities first, then merge both into `master`.

## TODOs
* [ ] Wait for #168 to be merge
* [x] Do not run on `latest` ESP-IDF until #180 is resolved.


martinberlin:
This is a very interesting contribution that can help a lot on all future additions. I will catch up as soon as I can hopefully next week.

martinberlin:
Hello @jdoubleu 
I would like to try this. I was just thinking that it would be cool also to first merge #168 that is kind of a big refactoring of how boards are organized done by Micke Prag. 

But still this is independent and with a bit of work it will adapt without problem to that changes, since the way to configure the project is still done by **menuconfig**. About this and the other PR #180 how do you suggest to proceed?

Just tell me what things need to be tested and in what order and I will try my best to organize it in the next days.

jdoubleu:
As you've mentioned, the CI will simply build all examples. At least it does `idf.py set-target esp32`, which should create an `sdkconfig` from the defaults. If that process didn't change, the CI should still work fine.

Nonetheless, I could create a duplicate branch off of this one and rebase it onto #168 to be certain it still works. It is then up to you what to merge first.

Unfortunately, the CI does not run for this PR, yet. In my fork it does, so I can test it there.

Ignore #180 for now, because there would be some breaking changes in there, that need to be addressed first.

jdoubleu:
I've removed the runs on the `latest` ESP-IDF version. They should be restored with #180.

As you can see at https://github.com/jdoubleu/epdiy/actions/runs/2299846098, everything builds successfully. I suggest to merge this PR first, then.

martinberlin (approved):
Looks good to me

PR #178, opened 2022-05-08
change boards_local.txt

Yardie-:
I have modified the lilygo section only. 
I have removed non lilygo screens (seemed pointless and potentially confusing).
I have also changed 
lilygo_t5_47.upload.maximum_size to 6553600
and
lilygo_t5_47.upload.maximum_data_size to 4521984
This sets the compile flag CONFIG_ARDUINO_ISR_IRAM
and as such the ARDUINO_ISR_FLAG is set to ESP_INTR_FLAG_IRAM 
which is 
#define ESP_INTR_FLAG_IRAM          (1<<10) ///< ISR can be called if cache is disabled

There are numerous warnings in the source of how this effects other code though. 
on a linux system
grep -r -A 2 ESP_INTR_FLAG_IRAM ~/.arduino15/packages/esp32/*
will show these. 
specific ones 

i2c.h: *             If the PSRAM is enabled and `intr_flag` is set to `ESP_INTR_FLAG_IRAM`, `data` should be allocated from internal RAM.

timer.h: *       If the intr_alloc_flags value ESP_INTR_FLAG_IRAM is set,
 the handler function must be declared with IRAM_ATTR attribute
 and can only call functions in IRAM or ROM. It cannot call other timer APIs.
 Use direct register access to configure timers from inside the ISR in this case.

This is a file that normally needs to be copied during the original setup of epdiy and as such this change will only occur if the user recopies the file. Somehow this needs to be documented for existing users.

### DIFF
--- a/boards.local.txt
+++ b/boards.local.txt
@@ -475,8 +475,8 @@ epdiy_v4.menu.DebugLevel.verbose.build.code_debug=5
 lilygo_t5_47.name=LILYGO T5-4.7 inch e-paper
 
 lilygo_t5_47.upload.tool=esp32:esptool_py
-lilygo_t5_47.upload.maximum_size=1310720
-lilygo_t5_47.upload.maximum_data_size=327680
+lilygo_t5_47.upload.maximum_size=6553600
+lilygo_t5_47.upload.maximum_data_size=4521984
 lilygo_t5_47.upload.wait_for_upload_port=true
 lilygo_t5_47.upload.flags=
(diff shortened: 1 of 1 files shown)


martinberlin (approved):
Looks good, but I would really like that someone that uses also arduino, tests this and gives feedback.

Yardie-:
Yeah @martinberlin  so would I. I will deal with the flack if there are any problems.

PR #179, opened 2022-05-08
add epd_powerdown for lilygo

Yardie-:
Add lilygo specific function that did my head in for a while.
When the power is turned off to the epd it seems the power to the touch screen is also disabled. Below are the notes I added.

On lilygo disable display power but not touch screen.
  The epd power flag was re-purposed as power enable
   however it also disables the touch.
  this workaround may still leave power on to epd
  and as such may cause other problems such as grey screen.
  please also use poweroff when you sleep the system
  wake on touch will still work just not the I2C interface.


martinberlin (approved):
Looks good, go ahead and merge

martinberlin:
Just add a small addition in the Readme please mentioning what is the use-case for this function.
I suggested Valentin that we start a Wiki to add specific documentation for different Boards / Epaper models but I guess that has to wait.

Yardie-:
Thanks @martinberlin. 
No worries.

Yardie-:
Ok that's done. Sorry I had to do it by hand and don't know markdown,

martinberlin (changes requested):
Please update the readme and remove the epd_poweroff_all() since I think is a function that does not exist anymore

src/epd_driver/include/epd_driver.h
```diff
@@ -206,6 +206,30 @@ void epd_deinit();
 /** Enable display power supply. */
 void epd_poweron();
 
+#if defined(CONFIG_EPD_BOARD_REVISION_LILYGO_T5_47
```
jdoubleu, line 209: There's a missing `)` here, that's why the CI currently keeps failing: https://github.com/vroland/epdiy/runs/6369702620?check_suite_focus=true

martinberlin:
This merge in a single commit: fd8a290f68d360c33ba7833487f38566e23e7178

 was reversed since it added a function **cfg_powerdown** is a exact copy of existing funcion **cfg_poweroff**
 
 CC @mickeprag

PR #181, opened 2022-05-10
use large partition layout for examples demo by default

jdoubleu:
Fixes build failure in CI: https://github.com/vroland/epdiy/runs/6368268549?check_suite_focus=true

The firmware size of the `examples/demo` slightly increased, so that it wouldn't fit into the default partition. I've selected the "large" partition layout and set it as default.

Must be, because of changes introduced in the meantime: https://github.com/vroland/epdiy/compare/f3feb5a93c2770260d46671a05e8341f3e6d06ae..c7bf7485f97abda61f74bff91cbaa769ac2f83b8. I guess the code changes increased the firmware by a bit.


martinberlin (approved):
All good

Issue #191, opened 2022-07-07
Assert when rendering a string with a space in it


dot4qu:
If either `epd_write_string` or `epd_write_default` are passed a string with a space in it, the call to `malloc` when drawing the glyphs will assert.

This is the line that causes the issue: https://github.com/vroland/epdiy/blob/master/src/epd_driver/font.c#L136

This only happens if the `CONFIG_HEAP_ABORT_WHEN_ALLOCATION_FAILS` sdkconfig option for the esp-idf SDK (located at `Component config -> Heap memory debugging -> Abort if memory allocation fails`) is set to Y, and the font in use is compressed. 

It's because a space has zero width/height in the font arrays, so the call to malloc is given a size of zero. Passing zero to malloc is undefined behavior, and the newlib version of malloc considers it's an error. If the heap memory debugging is turned on, this hard asserts the system.

I think a fix is just adding a check of `bitmap_size > 0` to the same if-statement below checking compression. i.e line 135 becoming:
```
  if (bitmap_size > 0 && font->compressed) {
```

At the very least, updating the README to specify that the menuconfig setting for heap memory debugging can lead to asserts. I'm happy to put up a PR with either option, just let me know the preferred way of handling it

Full crash stacktrace:
```
Memory allocation failed


Backtrace:0x4008238d:0x3ffbe4200x4009104d:0x3ffbe440 0x400d4b4f:0x3ffbe460 0x400825c2:0x3ffbe480 0x40098741:0x3ffbe4a0 0x400862d3:0x3ffbe4c0 0x400e25b6:0x3ffbe520 0x400e265c:0x3ffbe590 0x400e0a43:0x3ffbe5d0 0x400e0a8d:0x3ffbe620 0x400dcd29:0x3ffbe640 0x401959eb:0x3ffbe660 0x40094701:0x3ffbe690 
0x4008238d: panic_abort at /Users/<username>/Developer/esp/esp-idf/components/esp_system/panic.c:409

0x4009104d: esp_system_abort at /Users/<username>/Developer/esp/esp-idf/components/esp_system/esp_system.c:135

0x400d4b4f: heap_caps_alloc_failed at /Users/<username>/Developer/esp/esp-idf/components/heap/heap_caps.c:59

0x400825c2: heap_caps_malloc_default at /Users/<username>/Developer/esp/esp-idf/components/heap/heap_caps.c:191

0x40098741: malloc at /Users/<username>/Developer/esp/esp-idf/components/newlib/heap.c:24

[...]

PR #192, opened 2022-07-08
Don't try to malloc for a character glyph with size zero

dot4qu:
Fixes https://github.com/vroland/epdiy/issues/191

Tested by running on custom board and drawing text with spaces to screen.


martinberlin:
Merged, thanks for your contribution @dot4qu

PR #201, opened 2022-08-26
Place tinfl_decompressor in heap

mickeprag:
This object is large (~11 KB)) and is taking up a much of the ram of the ESP32.
Uncompressing a font is not a continous operation and this object is not
needed all the time so placing this temporarily in heap when needed makes more
sense.

### DIFF
--- a/src/epd_driver/font.c
+++ b/src/epd_driver/font.c
@@ -33,11 +33,6 @@ static utf_t *utf[] = {
     &(utf_t){0},
 };
 
-/**
- * static decompressor object for compressed fonts.
- */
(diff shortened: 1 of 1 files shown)


mickeprag:
With the decompressor in static ram I am having issues with opening a ssl connections since I have no memory. Moving this to heap solves the memory issue since this will free up a lot of ram.

martinberlin:
Thanks @mickeprag for your contribution.

For me is a good update. Can you review this and merge it @vroland please?
If you have no time please tell me and I will do it.

PR #218, opened 2023-02-21
Add screen diagnostics example app

jdoubleu:
This example app implements a simple interactive shell that lets you tinker with the display. You can draw simple shapes (i.e. everything the _epd driver_ provides), read system information, etc. I've ported some of my screen test algorithms (i.e. `render_stairs` and `render_grid`) as well.

This can be helpful when debugging incompatible or broken displays or just experimenting with the draw functions.

Also note: the `power_on` command basically obsoletes the [calibration_helper](https://github.com/vroland/epdiy/tree/master/examples/calibration_helper) example.


vroland:
Thank you, that looks useful! I'll test it a bit and hopefully merge it soon :)

vroland:
Wow, that's awesome, I really like the interactive console! 
I just noticed one minor thing: When I try to get the temperature without enabling power, the I2S read fails and the ESP runs into an assert and reboots. However, for some reason, it then goes into a boot loop trying to do the read again when it reboots, even after power cycling. Do you know why that is? I could only break the loop by adding a check to the temperature reading function, which I should probably do anyway...

jdoubleu:
> When I try to get the temperature without enabling power, the I2S read fails and the ESP runs into an assert and reboots. [...]

Hm, that's interesting. How do you know, it tries to read the temperature again? No, I have no clue why this happens.

> [...] by adding a check to the temperature reading function [...]

That's probably a good idea! Gonna add this check.

jdoubleu:
Sorry for the long delay.

> I just noticed one minor thing: When I try to get the temperature without enabling power, the I2S read fails and the ESP runs into an assert and reboots.

I am unable to reproduce this issue. With my v5 board, reading the temperature, even if the board isn't manually `epd_poweron()`'d seems to work fine.

Nonetheless I've added a fix which always turns on and off the epd before reading the temp.

martinberlin:
This weekend I will start testing this that I think is a great contribution. @jdoubleu could you please merge back master that is now updated into your pull request?
I think won’t be much effort to update this since only initialization of epdiy changed and it’s not anymore per menuconfig.
So just a few lines need to be updated.

vroland:
Hey @jdoubleu , I created a branch where your changes are rebased on epdiy 2.0.0. If you care that we merge your branch, could you rebase on `main` and cherry-pick https://github.com/vroland/epdiy/commit/dfeee32262552503a71e4f6439e422a1950a0fed? 
I can also just merge that branch, if you don't care about having specifically the branch from your fork merged.

vroland:
Nvm, I can actually just merge this and then apply the fix :D Thanks again for the nice contribution, very useful!

Issue #220, opened 2023-03-01
GPIO12 unusable


Fabian-Schmidt:
Hi,
I am using GPIO pin 12 as an UART pin and the library disables it every time on powerdown of the display.

https://github.com/vroland/epdiy/blob/12c67cead7a76e90f692fa2c3afcda87831badd0/src/epd_driver/i2s_data_bus.c#L168

Can this behavior be changed? I understand the power savings desire. Is it possible to add at least a `#if` directive to disable this behaviour?

Thanks,
Fabian

mickeprag:
Is this a custom board or the official?
For a custom board I would suggest you add your own board implementation in your project. Copy from epdiy and change whatever you need.

Fabian-Schmidt:
It is a Lilygo T5-4.7 board.

Fabian-Schmidt:
I am using a Lilygo T5-4.7 Board which is supported by the library and therefore I would like to avoid forking the library. I did fork it to validate that removing this line will fix my issue, but I would rater like to keep it together in your branch.

The problem is that the Lilygo T5 device has so few ports (4) and I need unfortunately all of them.

mickeprag:
You do not need to fork the library. It is possible to supply your own board implementation and still use the mainline library.

martinberlin:
@Fabian-Schmidt this IO is not used in any PCB version.
You can just make a merge request removing this line and we can review it testing in before merging it. IMHO don't think that this will affect consumption at all.
Only note that I'm aware of in the datasheet about this IO is this one:

> GPIO12 is internally pulled high in the module and is not recommended for use as a touch pin.

So keep that in mind in your design, if it's pulled-up high, then reading that IO should return a 1 by default. Give it a try.

martinberlin:
On hold till https://github.com/vroland/epdiy/pull/251 is merged

martinberlin:
Hello Fabian, can you make the pull-request but from the current **main** branch please?
Thanks!

martinberlin:
Closing here since the request was merged

PR #222, opened 2023-03-04
Make GPIO 12 usable

Fabian-Schmidt:
The epdiy library disables GPIO 12 on power down/off operation.
Despite GPIO PIN 12 is not being used by any supported board for any operation within this library.

My supported LILYGO® T5-4.7 has only 4 free GPIO (12,13,14,15 - in two connectors on the upper left) and I need all of them.
LILYGO® T5-4.7

GPIO 12 has limited usage as it is a strapping pin. I am using it as a UART Tx pin and it is working fine. (Cannot use it as UART Rx pin - if pulled high during boot ESP does not detect it has PSRAM).

This PR fixes issue #220


martinberlin:
V6 PCB exposes IO12 as external IO
I think removing the GPIO12 isolate is OK but if I'm missing something please let me know @vroland @jdoubleu 
What I'm not sure if it's good to do it like this or is better to add a #ifdef constraint and do it only for Lilygo EPD47 board. Please clarify this point.

Fabian-Schmidt:
I can alter the PR to add an `#ifdef` so by default it is disabled, but if somebody needs GPIO12 they can use it.

vroland:
Could you maybe add the code that is deleted to the docs, mentioning that it can be added to optimize deep sleep consumption?

Fabian-Schmidt:
Hi @vroland, I added to deleted code to the readme.

martinberlin (approved):
I think this is fine. I would mention that the low power for that gpio is only in ESP32 since this is not an issue in new ESP32-S3 board

martinberlin:
Sorry @Fabian-Schmidt lets close this MR and add this small change when https://github.com/vroland/epdiy/pull/251 is merged since the code changes significantly after that is merged

PR #248, opened 2023-05-03
Align dma buffers

mickeprag:
The size of the buffers used by DMA must be word aligned. If the width of the display is not aligned we increase the buffer slightly.

For instance the display ED060KC1 has a width of 1448 pixels. The DMA buffer would be 362 bytes (1448/4) and this will not work with the ESP32 DMA driver. Instead the buffer will be increased to 364 bytes.

Also a small fix for faulty boards which made the power on code loop forever.


vroland:
Looks good to me, thank you!

PR #250, opened 2023-06-03
Make it possible to set the display type at runtime

mickeprag:
The goal is to be able to support different displays at runtime. I have a use case where one firmware should support more than one display.

This change is a non breaking and backwards compatible addition. It will still be able to select a fixed display using `menuconfig`. If no display is selected the code must call `epd_set_display()` before `epd_init()`, similar to selecting the board at runtime.

### DIFF
--- a/src/epd_driver/CMakeLists.txt
+++ b/src/epd_driver/CMakeLists.txt
@@ -1,5 +1,6 @@
 
 set(app_sources "epd_driver.c"
+                "epd_display.c"
                 "epd_board.c"
                 "render.c"
                 "display_ops.c"
--- a/src/epd_driver/Kconfig
+++ b/src/epd_driver/Kconfig
@@ -22,7 +22,7 @@ menu "E-Paper Driver"
 
         config EPD_DISPLAY_TYPE_ED133UT2
             bool "ED133UT2"
-        
+
         config EPD_DISPLAY_TYPE_ED047TC1
             bool "ED047TC1 LILYGO 4.7 inch, correct grays"
         config EPD_DISPLAY_TYPE_ED047TC2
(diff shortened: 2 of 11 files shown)


martinberlin:
This looks great to me for current version. 
Only issue is that as you've already seen @vroland started already with a similar idea in s3_lcd branch.
Which I suppose will be the next V7 version of Epdiy PCB. Now I'm not really sure how this will be planned in order to support old versions. But it will be nice for him also to review this PR since maybe it can take more months to see the light.
Thanks a lot Micke for your always thoughtful contributions to this project

mickeprag:
Yes, this is why I opened this PR now. Feels not good that we both are trying to solve the same thing twice. Instead we should try to merge mine and @vroland work to find a solution he is also fine with.

vroland:
@mickeprag Thanks for starting on this! I know that working on the v7 branch quietly right now is not optimal, maybe I should open a draft PR to at least make it more visible. Would you mind if I just cherry-pick your changes and integrate them into this branch? I'll change the display configuration a bit to also include bus width and bus speed for v7, but already having all the places in the draw functions changed is quite convenient :)
`lut.c` will need some care to make sure the lookup functions still optimize well if the width is not known compile-time, but that is possible with some annotations I added on the v7 branch.

mickeprag:
Sure, that sounds like a good idea!

>  I'll change the display configuration a bit to also include bus width and bus speed for v7

Sounds good!

martinberlin:
Great as I can see now this is already done in latest updates in V7 pull request #251
Micke can you take a look and confirm so we close here? This is the way that we instantiate the display now:

```C
  epd_init(&epd_board_v6, &ED060XC3, EPD_LUT_64K);
  // Set VCOM for boards that allow to set this in software (in mV).
  // This will print an error if unsupported. In this case,
  // set VCOM using the hardware potentiometer and delete this line.
  epd_set_vcom(1560);
  hl = epd_hl_init(WAVEFORM);
```

mickeprag:
This is a breaking change but it is ok with me.

PR #251, opened 2023-06-12
Tracking PR: epdiy v7 and dynamic display configuration

vroland:
This PR tracks V7 hardware development and software updates. List of tasks:

- [x] V7 hardware prototype
- [x] V7 final hardware designed
- [x] V7 final hardware tested
- [x] Deepsleep current verified (~30uA for now, may be enhanced by isolating LCD pins)
- [x] Software support for V7 and previous versions (Currently, only V7 works with this version)
- [x] dynamic display selection


src/epd_driver/CMakeLists.txt
martinberlin, line 1: This is looking good. Will start testing this in the V7 board very soon

martinberlin:
Just tested this with a V7 board and this configuration and the examples/demo program.

The beginning with the "Loading demo" scroll-bar runs fine (A bit slow but OK)

```c
// Using EPD_LUT_1K it hangs directly at the beginning
  epd_init(&epd_board_v7, &ED097TC2, EPD_LUT_64K);
  // Set VCOM for boards that allow to set this in software (in mV).
  epd_set_vcom(1560);
```

Then when it's trying to load the picture it hangs with this error:

```
I (21791) epdiy: calculating diff..
I (22041) epdiy: highlevel diff area: x: 354, y: 25, w: 492, h: 768
I (22051) epdiy: starting update, phases: 30
W (22061) epd_lcd: draw frame draw initiated, but an error flag is set: 400
``` 

Video https://github.com/vroland/epdiy/assets/2692928/494b842a-b795-45ea-a8dc-08d916502f61

martinberlin:
@vroland with this last 2 commits we can already test this examples in IDF 4.4 / Arduino ?
Will make some tests over the weekend with my 9.7" and V7 board.

vroland:
It does compile in arduino, but for some reason I have to use much slower bus speeds for it to work consistently, probably caused by some stuff arduino is running in the background. I wasted an evening on this already, maybe I can reproduce it by compiling arduino as a component but that will take even more time... :/

src/output_lcd/lcd_driver.c
```diff
@@ -381,7 +381,17 @@ static esp_err_t s3_lcd_configure_gpio()
     return ESP_OK;
 }
 
-void epd_lcd_init(const LcdEpdConfig_t* config, int display_width, int display_height) {
+void IRAM_ATTR epd_lcd_init(const LcdEpdConfig_t* config, int display_width, int display_height) {
+
+    if (CONFIG_ESP32S3_DATA_CACHE_LINE_SIZE < 64) {
```
martinberlin, line 386: This looks very useful to me. Instead of the cache override I would mention that displays.c bus_speed should be lower to work with a line size <64 bytes

martinberlin:
@vroland it looks awesome. Tomorrow will check some of the examples with IDF 5.1 see if I can help fixing them so they pass the tests.
What is the MPD status one? That I never tried. Maybe it will be nice to add a short 4 lines Readme to explain what it does.

vroland:
It's a client for the music player daemon, to show album cover and current playlist. I added a readme :)

martinberlin:
I've been testing the examples specially to give you a hand to fix mine, this is just that IDF v5 is picky about the integer type in any kind of logging:

```
/app/vroland/epdiy/examples/www-image/main/jpg-render.c:517:12: error: format '%lu' expects argument of type 'long unsigned int', but argument 2 has type 'size_t' {aka 'unsigned int'} [-Werror=format=]
  517 |     printf("Free heap after buffers allocation: %" PRIu32 "\n", xPortGetFreeHeapSize());
  ```

**Lange Rede kurzer Sinn** :) You can just copy the fixed examples from my branch [s3_color_research](https://github.com/vroland/epdiy/tree/s3_color_research/examples/www-image) where they are fixed. 
Then I will go throw them and try them in my 9.7" epaper. 

Maybe it's a good idea to add a WiKi page mentioning in a short and straight-forward way that all examples except Lilygo EPD47 are meant for a 9.7" parallel epaper and if using a different model, they have to be edited, since most people is at this point accustomed to the **menuconfig** option and after this is merged, that will be not any more the way to set this up.
Is a breaking change, but is nicer to do it this way, and it will be also nicer to the Platformio users since they do not have to use constants anymore in their platformio.ini configuration file.

After this is fixed, only the MPD status and terminal examples should be fixed, for both IDF 4.4 and 5.

**Comment by vroland** [comment]
@martinberlin Thanks, I used the fixes and did some more (allocating a smaller buffer for the downloaded image, updating the cert).

**Comment by martinberlin** [comment]
> @martinberlin Thanks, I used the fixes and did some more (allocating a smaller buffer for the downloaded image, updating the cert).

Awesome great idea to allocate the buffer using content length header! Now the www-image examples pass the tests.

**Comment by martinberlin** [comment]
@vroland just check this error for the weather example

>  esp32-arduino requires CONFIG_FREERTOS_HZ=1000 (currently 100)

and found this mention in the [Platformio ESP32 repository](https://github.com/platformio/platform-espressif32/issues/846)
Is there a specific config that should be added when using the Arduino repository?

I know that this example is quite specific and has some author references but maybe one day we should do our own weather example that is easier to setup (Out topic for the PR of course!)

File: src/epdiy.c [hunk 2]
```diff
@@ -532,3 +534,12 @@ int epd_width() {
 int epd_height() {
     return display->height;
 }
+
+void epd_set_lcd_pixel_clock_MHz(int frequency) {
+#ifdef RENDER_METHOD_LCD
+    void epd_lcd_set_pixel_clock_MHz(int frequency);
```
martinberlin, line 540: This line is not defined already in the header?
Or maybe it’s me that I’m too sleepy and should already stop checking code :P

vroland: It's not quite the same, `epd_lcd_set` vs `epd_set_lcd` ;) One is the internal LCD method, for which I didn't want to include the header so I reproduced the function declaration.

idf_component.yml
```diff
@@ -0,0 +1,4 @@
+version: "2.0.0"
+description: "Drive parallel e-Paper displays with epdiy-based boards."
+url: "https://github.com/vroland/epdiy"
+license: LGPLv3
```
martinberlin, line 4: Can you please check the comment I made in the PR?
In the Espressif docs mentions: license: The license of the component. It should be a valid SPDX license identifier listed in https://spdx.org/licenses/ This field is optional. The license will be automatically detected if it’s not specified.
LGPLv3 does not seem to be listed in SPDX so leaving this field as is will drop a warning on any user trying to use IDF component manager

vroland: should be fixed now.

martinberlin (approved):
Awesome job!

martinberlin:
@vroland one additional request before merging this in master:
https://github.com/vroland/epdiy/releases Maybe is wise to make another minor release? Like **1.0.2**
Then if someone wants to still use older version in a firmware can just tag it like that. Otherwise we will force many Firmware developers using the component to update the integration in their code.

mickeprag:
Yes, I agree to do a release before any breaking changes. Also minor fixes could later also be ported to the 1.0 branch if needed.

vroland:
Of course, that makes a lot of sense :)

vroland:
I'm just missing some updates to the docs and readme, then this can be merged.

PR #263, opened 2023-11-15
fix CKV drift and frame restart in LCD driver

vroland:
This PR fixes an issue where the CKV clock would drift for 8 bit screen due to the use of dummy cycles which work around an issue in the ESP32 LCD FIFO. Additionally, dummy cycles are now also used for 16-bit parallel displays, fixing a pixel offset issue.

### DIFF
--- a/examples/screen_diag/main/epd.c
+++ b/examples/screen_diag/main/epd.c
@@ -17,7 +17,7 @@ static int s_temperature;
 
 void initialize_screen(void)
 {
-    epd_init(&DEMO_BOARD, &ED097TC2, EPD_LUT_64K);
+    epd_init(&DEMO_BOARD, &ED060XC3, EPD_LUT_64K);
     // Set VCOM for boards that allow to set this in software (in mV).
     // This will print an error if unsupported. In this case,
     // set VCOM using the hardware potentiometer and delete this line.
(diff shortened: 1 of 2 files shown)


vroland:
@martinberlin If you have the time, can you check if this fixes your issues with the color display?

martinberlin:
Hello Valentin,
I can confirm this works now perfectly with color. 

For some reason still out of my reach after merging this into s3_color_implementation branch now I need to swap red for blue. But the colors work correctly without the jump in row 1000. Great fix thanks a lot!

```
-107,13 +107,13 @@ uint8_t epd_get_panel_color(int x, int y, uint8_t r, uint8_t g, uint8_t b) {
     switch (c)
     {
       case 0:
-        return gamme_curve[r]; // R
+        return gamme_curve[g]; // Green
         break;
       case 1:
-        return gamme_curve[b]; // B
+        return gamme_curve[r]; // Red
         break;
       default:
-        return gamme_curve[g];
+        return gamme_curve[b];
     }
 }
```

martinberlin (approved):

vroland:
This PR changes also fixes the line start which may shift your color arrangement. But good to hear that it works :+1:

Issue #203, opened 2022-08-29
Adding Lilygo ESP32S3 Platform Support


THSLP13:
### It uses ESP32S3(N16R8) as controller,and uses ED047TC1(Soft ver.) as display.
I try to make a patch to my own code,but arduino IDE only throws a lot of(99+) error to me.
When I simply compile code from your example,IDE throws errors again,and throws the same errors.
So I write the issue.

The code from Xinyuan-LilyGO works normal and fully functional,but they hasn't update their code such as update mode,they just update the pinmap and I2S Sources.
(Hint:Select S3 branch to get S3 version driver from lilyGo)

Most errors contains declaration error.
Please notice me if you need further information about this problem.

martinberlin:
Sorry but you started wrong, because there is a PR that is still not merged, that will enable EPDiy to compile in IDF v 5.0
Until that is merged, you should start on that branch, instead of EPDiy master. 
I was going to make my own PR to add this board but if you manage to do it, go ahead.

martinberlin:
@THSLP13 since it's on your own best interest to have this working on Arduino, please help me test this PR:

https://github.com/vroland/epdiy/pull/180#issuecomment-1230878963
So you can give some feedback about it. Just get that PR branch and use it to compile your EPDiy library on Arduino IDE.

THSLP13:
> @THSLP13 since it's on your own best interest to have this working on Arduino, please help me test this PR:
> 
> [#180 (comment)](https://github.com/vroland/epdiy/pull/180#issuecomment-1230878963) So you can give some feedback about it. Just get that PR branch and use it to compile your EPDiy library on Arduino IDE.

Hi,thanks for your reply.
Due to my coding skill level I couldn't give much advice,but I will do my best to test the PR.

martinberlin:
When my Lilygo S3 arrives I will try to port the updates over the existing working branch on IDF v5. 

I started on my fork but is very raw still and needs to be debugged.

vroland:
I'm curious how the S3-based board will perform. I have some normal S3 dev boards on the way since the Lilygo board is a bit expensive and not flexible enough for prototyping, but I'm confident @martinberlin can make the necessary changes :)
Since we're planning an S3-based epdiy for the future, this will be good ground work anyway.

martinberlin:
@THSLP13 you can try this:

https://github.com/martinberlin/epdiy-rotation/tree/lilygos3

[...]

PR #272, opened 2023-12-25
Add #203 Lilygo S3 support

martinberlin:
This got more complex that I expected. Only because it's a mix of v7 (LCD module) and v2 that is where LIlygo started to implement their display.
Then the issue is that LEH, Latch enable signal, that is currently implemented directly in the LCD module (hsync) please correct me if I'm saying things out of the blue after my Christmas hang-over....it's still implemented using RTC in LIlygo S3 parallel solution.
So technically this is a mix of v2 and v7 boards, that is not simply copying v7 and updating some lines. 
https://github.com/martinberlin/epdiy-rotation/tree/lilygos3 -> There is my working solution to implement this but I feel going backwards implementing this "as is" here. @vroland or @mickeprag maybe I will need your wise advice


martinberlin:
Won't work on this anymore. Closing PR

Issue #300, opened 2024-04-24
epd_renderer_init failing on v7 board


bgrassy:
Hello! I recently got a v7 board from JLCPCB. I've been able to get things running on the v5 board before, but I'm running into issues with the dragon example project with this new board. I installed IDF version 5.2.1 through the VSCode extension, and was able to successfully build and flash the project to my board. The only change I made in the code was to adjusts the VCOM in `epd_set_vcom`. When monitoring after flashing, I see the following message repeatedly in the logs:

```
Rebooting...
�ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x9 (SPI_FAST_FLASH_BOOT)
Saved PC:0x403754a9
0x403754a9: esp_restart_noos at /home/brock/esp/v5.2.1/esp-idf/components/esp_system/port/soc/esp32s3/system_internal.c:159

SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3820,len:0x1568
load:0x403c9700,len:0x4
load:0x403c9704,len:0xe3c
load:0x403cc700,len:0x2f88
entry 0x403c992c
I (31) boot: ESP-IDF v5.2.1-dirty 2nd stage bootloader
I (31) boot: compile time Apr 24 2024 00:34:02
I (31) boot: Multicore bootloader
I (35) boot: chip revision: v0.2
I (39) qio_mode: Enabling default flash chip QIO
I (44) boot.esp32s3: Boot SPI Speed : 80MHz
I (49) boot.esp32s3: SPI Mode       : QIO
I (53) boot.esp32s3: SPI Flash Size : 2MB
I (58) boot: Enabling RNG early entropy source...
I (64) boot: Partition Table:
I (67) boot: ## Label            Usage          Type ST Offset   Length
I (74) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (82) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (89) boot:  2 factory          factory app      00 00 00010000 00177000
I (97) boot: End of partition table
I (101) esp_image: segment 0: paddr=00010020 vaddr=3c020020 size=85b28h (547624) map
I (192) esp_image: segment 1: paddr=00095b50 vaddr=3fc93300 size=0240ch (  9228) load
I (194) esp_image: segment 2: paddr=00097f64 vaddr=40374000 size=080b4h ( 32948) load
[...]

Issue #285, opened 2024-03-11
Prepare v7 lcd_driver.c to be ready for IDF 5.3 


martinberlin:
When trying to build this component with IDF 5.3 we came upon this errors (5.1 works correct)

### 1st error: lcd_ll_set_data_width

```
src/output_lcd/lcd_driver.c:501:5: error: implicit declaration of function 'lcd_ll_set_data_width'; did you mean 'lcd_ll_set_data_wire_width'?
```

This one I don't know why it comes but in the documentation of [ESP32S3 I see only a reference to data_width](https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/lcd.html?highlight=lcd%20module#_CPPv4N26esp_lcd_rgb_panel_config_t10data_widthE) but I don't find this function. Anyways it compiles correctly in IDF 5.1, and it of course exists:

_$ grep -rne lcd_ll_set_data_width ._

```
./components/esp_lcd/src/esp_lcd_panel_rgb.c:503:    lcd_ll_set_data_width(rgb_panel->hal.dev, rgb_panel->data_width);
./components/hal/esp32s3/include/hal/lcd_ll.h:277:static inline void lcd_ll_set_data_width(lcd_cam_dev_t *dev, uint32_t width)
```

### 2nd error: Missing argument to function 'lcd_hal_cal_pclk_freq'

```
src/output_lcd/lcd_driver.c:546:21: error: too few arguments to function 'lcd_hal_cal_pclk_freq'
  546 |     uint32_t freq = lcd_hal_cal_pclk_freq(&lcd.hal, 240000000, lcd.config.pixel_clock, flags);
      |                     ^~~~~~~~~
```

This happens because in idf 5.3 there is a new argument at the end called hal_utils_clk_div_t* **lcd_clk_div**

Reference in [5.3 LCD HAL](https://github.com/espressif/esp-idf/blob/60a2bf6a68867fd468a250966230a067f8d97ea2/components/hal/lcd_hal.c#L16).

vroland:
Good to know, unfortunately the HAL is not guaranteed to be stable, but the high-level drivers don't work for our purposes :/

martinberlin:
Interesting I didn’t know that.
Just wanted to mention that in 5.2 version of IDF there are also issues https://github.com/vroland/epdiy/issues/300 so we should prepare to handle those cases because many people will fall there @vroland 
[...]

PR #307, opened 2024-04-30
fix render task priority value

vroland:
The current value was outside of the legal limits by one. Since 5.2 this was enforced causing us to hit an assert.


vroland:
fixes #300 .

martinberlin (approved):
Looking good!
Let's see if we can fix this for #285 also!

PR #314, opened 2024-05-25
Add clang-format make targets and CI checks

vroland:
Adds a .clang-format definition and checks in CI to enforce consistent code formatting going forward.


mickeprag:
👍From me! 😎

src/board/epd_board.c
```diff
@@ -1,11 +1,10 @@
 #include "epd_board.h"
 
-#include <stddef.h>
 #include <esp_log.h>
+#include <stddef.h>
```
martinberlin, line 4: Just curious, why the order of the includes matter?

martinberlin (approved):
I second Micke!
Looks good for me except for the include ordering because I don’t see the point of it.

src/board/epd_board.c
schuhumi: I think the formatter does this for style reasons. In `.clang-format` the option `BasedOnStyle: chromium` is set. If you follow the [docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#basedonstyle) you'll eventually end up at [Google's cpp styleguide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes), which states:

> Within each section the includes should be ordered alphabetically.

Interestingly, I think this can become a problem. Namely, FreeRTOS requires to `#include <freertos/FreeRTOS.h>` before any other FreeRTOS components. But `#include <freertos/atomic.h>` for example would alphabetically come before it, and that header would then not let you compile:

```c
#ifndef INC_FREERTOS_H
    #error "include FreeRTOS.h must appear in source files before include atomic.h"
#endif
```

martinberlin: Thanks for the explanation. I don’t like that automatic organization of includes. IMHO this should be completely turned off

vroland: The problems with include order on FreeRTOS can be solved by separating blocks of imports with a newline. However, I also don't see the value of sorted includes, so I turned it off.

PR #353, opened 2024-09-01
Refactor and test LUT functions

### LINKED ISSUE #284: MODE_PACKING_2PPB renders "stripes"
When using MODE_PACKING_2PPB 

`epd_draw_base(epd_full_screen(), state->front_fb, epd_full_screen(), MODE_PACKING_2PPB | PREVIOUSLY_WHITE | MODE_GC16, temperature, NULL, state->waveform);`

instead of MODE_PACKING_1PPB_DIFFERENCE where diff_area = epd_difference_image_cropped(...)

`epd_draw_base(epd_full_screen(), state->difference_fb, diff_area, MODE_PACKING_1PPB_DIFFERENCE | MODE_GC16, temperature, state->dirty_lines, state->waveform);
`

the front_fb is printed in "stripes" on the eink, see attached image.

IMG_6153

ED133UT2
epdiy v7 esp32s3

vroland:
Refactors waveform lookup calculations to be more maintainable and adds tests to avoid regressions. In the process, line masking is simplified and ordering problems in 2ppB mode are fixed (this should fix #284).
8ppB now supports both PREVIOUSLY_BLACK and PREVIOUSLY_WHITE.


martinberlin:
Looks very nice Valentin. Just tell me when it’s good to step in and test it

martinberlin:
I will be testing this over the weekend. Question @vroland 

1. In [line_queue.c](https://github.com/vroland/epdiy/pull/353/files#diff-9a7dcc55f94fea7a232a01cfebfd656b668290bb73ce3ced587a7f9ffc04ef09) basically the masking has been removed right, now is handled different. Can you explain in a few work how it changed?

2.  How can we use the new [fb_mode_test](https://github.com/vroland/epdiy/tree/3fd938481bd0965a1b186f081c70928cd30e6999/examples/fb_mode_test) ?
Visually it draws something on the display. And based on that we can check what mode would work best and also make sure that the display works correctly?

3. In the process, line masking is simplified and ordering problems in 2ppB mode are fixed (this should fix https://github.com/vroland/epdiy/issues/284)
That's awesome. 

Is there now a way to also draw something in 8ppB mode, basically full monochrome, using the high level API?
Meaning we can have a smaller framebuffer and make UX faster than using the 2ppB mode. That could be interesting for some users testing LVGL.

martinberlin (approved):
All examples seem to still work after this upgrade. Would like to have the questions made before replied but does not have to be in this PR.
We can move them to discussions.

j4k32:
Works perfectly, thank you!

vroland:
For 1: The issue was that on the esp32-s3 the line queue was used for output lines after the waveform lookup, but for v6 and below it was used for raw input data. I moved the masking so it always operates on the output lines.

For 2: The idea is to have a quick visual test that things look correct with the different buffer settings. Maybe eventually we can also use it to test the actual output in the simulator.

The last question: With the high level API, you can't use full monochrome because the diffing and draw functions are not implemented for it. You'd need to use the low level API :/

PR #368, opened 2024-11-02
Example: HTTP Server

BackSlasher:
Runs an HTTP server that will draw images on the screen. Useful for setting up a small digital frame that can be remotely controlled.
Has a friendly Python client,

Video:
https://github.com/user-attachments/assets/dd4d095b-7b5a-4d57-b403-6f1afc7d6c81


BackSlasher:
I'll fix the formatting

BackSlasher:
I think now we should be good, not sure why the single format error still shows up - it's clean on my laptop, and I'm not sure what to change it to

martinberlin:
Cool example. It will be really nice if you could add also a:

**/drawjpg** endpoint, that will receive the JPG file, decompress it and draw it on the display 

but knowing that you make a Python Client is also ok. It’s just that this way could be easier for someone to send a JPG image from anywhere including a simple browser in the browser of any mobile device.
I’ve did 2 examples of JPG decompression that you can see in www-image one eses the ROM jpegd and the other is faster and uses Larry Bank JPEGDEC component. In the v7 there is a enough PSRAM to upload and decompress a quite fat image.

BackSlasher:
Thanks for the idea, I'll have a look. If adding JPG support will be easy, I'll do it

BackSlasher:
On a second thought, would you mind merging it now, and we can add the jpg endpoint later?
I think that having this server without jpg is better than not having it at all, and we can always improve the interface later

vroland:
Looks interesting! I'll have a closer look and try it out hopefully later this week :)

martinberlin:
A bit of early feedback from my side here:

1. 

If you clone this, switch to http-server branch and try to build it

```
undefined reference to `epd_lcd_deinit'
```

This is because is not taking **esp32s3** as default target. **sdkconfig.defaults** should point to ESP32S3 target that is the last PCB.  My advice, just copy: sdkconfig.defaults.esp32s3  to sdkconfig.defaults so the user can directly compile this for the last v7 PCB.

2.  two configuration files:

    In main/server.h, edit WIFI_SSID and WIFI_PASSWORD to match your wifi config
    In main/main.c, edit n_epd_setup to refer to the right EPD screen

This could be ideally reduced to one settings.h header file or just one of them, where you can edit a single file and just build. 
For me this would be beneficial for 1st time users of this example. You know as less complicated is better so there is more people getting it right from the 1st time.

examples/http-server/main/main.c
```diff
@@ -0,0 +1,147 @@
...
+        sprintf(response, "Missing header height");
+        httpd_resp_set_type(req, "text/plain");
+        httpd_resp_set_status(req, "400");
+        httpd_resp_send(req, response, HTTPD_RESP_USE_STRLEN);
+        return ESP_OK;
+    }
+
+    // READING STREAM
+    int req_size = req->content_len;
+    char* content = (char*)heap_caps_malloc(req_size, MALLOC_CAP_SPIRAM);
```
jonasschnelli, line 77: This memory buffer is not deallocated after receiving data, which could cause a memory leak.

examples/http-server/main/main.c
```diff
@@ -0,0 +1,147 @@
...
+    }
+    int current_pos = 0;
+    int amount_recieved;
+    while ((amount_recieved = httpd_req_recv(req, (content + current_pos), req_size)) > 0) {
+        ESP_LOGI(__FUNCTION__, "Read %d bytes\n", amount_recieved);
+        current_pos += amount_recieved;
+    }
+    if (amount_recieved < 0) {
+        char msg[50];
+        free(content);
```
jonasschnelli, line 92: maybe `heap_caps_free()` instead?

martinberlin (changes requested):
If possible check my comments and the mentions done by another reviewers of the code

BackSlasher:
thank you all for the feedback!

examples/http-server/send_image.py
```diff
@@ -0,0 +1,100 @@
...
+class Dimensions(NamedTuple):
+    width: int
+    height: int
+
+def info(hostname):
+    resp = requests.get(f"http://{hostname}")
+    resp.raise_for_status()
+    return EpdInfo.from_response(resp)
+
+import PIL
```
vroland, line 43: Just a nitpick: I'd put all the imports to the top.

BackSlasher: You're right

martinberlin:
hello @BackSlasher please do not force commit like this because is hard to follow history. Usually you force commit something when you want to go back to a certain point but not when you go forward correcting stuff.

BackSlasher:
I have no problem adapting to what's convenient to you :)
I personally find squashing (and as a result, force-pushing) to keep PRs to a single commit. From now on for PRs on this repo I will avoid this.

martinberlin (approved):
Looks good to me. @vroland if you want to try it go ahead, I like the example, and maybe at some point will contribute adding pure JPG decoding ability

jonasschnelli:
I have plans to test this further as soon as my boards arrive.

vroland:
Regarding the force-pushing: I don't have a problem with this, do whatever you want in the PR; I squash before merge anyway ;)

PR #369, opened 2024-11-04
Fix ED078KC1 screen resolution and add PPI to the overview page

jonasschnelli:
Ready to remove PPI calculations if they're not desired on the overview page


vroland:
Thanks!

vroland (approved):

PR #376, opened 2024-11-21
Remove sdkconfig defaults from httpserver example

jonasschnelli:
Couldn't figure out why `idf.py build` always built for ESP32S3 instead of ESP32.
Other examples have no default too.


jonasschnelli:
The esp32s3 file already exists (see https://github.com/vroland/epdiy/blob/main/examples/http-server/sdkconfig.defaults.esp32s3). This PR removes only the unnecessary `sdkconfig.defaults` file (not the `sdkconfig.defaults.esp32s3`).

vroland:
Ah right, makes sense :sweat_smile:

vroland (approved):

PR #385, opened 2024-12-17
Adding new supported display ED052TC4 1280x780

martinberlin:
Adding this nice and cheap display (17€ in Aliexpress)
Similar to Lilygo EPD47 but slightly bigger and more resolution. Grays work but only 9 scales or so, probably needs better gray matrix. I will stop calling it waveforms since it's actually a 2 bit data stream, with transitions from Gray1 to Gray2 :D
More grays but still not perfect using the Lilygo ED0457TC2 double passes waveform file.

ADAPTER Kicad files:
https://github.com/martinberlin/epdiy-hardware-versions/tree/main/adapters/40_to_WP27D-P050VA3

IMG_3218
camphoto_126398554 rotated

In this update I'm also adding some fixes to my www-image example:

- Moving display into settings.h so you don't have to touch the main file
- Updating logging related errors (The usual this is not an (int) or not a large int)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Displays
 |ED047TC1|4.7"|960 x 540<br>234 PPI|yes, tested|40-pin|40|LILYGO 4.7" EPD|Supported only by 4.7" e-paper board by LILYGO
 | ED050SC5 | 5" | 600 x 800<br>200 PPI | yes, tested       | THD0510-33CL-GF | 33 | v5 |
 | ED050SC3 | 5" | 600 x 800<br>200 PPI | yes (should work) | THD0510-33CL-GF | 33 | v5 |
+| ED052TC4 | 5.2" | 1280 x 780<br>??? PPI | yes (should work) | WP27D-P050VA3 | 50 | v5 |
 | ED133UT2 | 13.3" | 1600 x 1200<br>150 PPI | yes, tested | adapter board | 39 | v2 | Adapter Board required, also PENG133D
 | ED060XC3 | 6" | 758 x 1024<br>212 PPI | yes, tested | THD0515-34CL-SN | 34 | v5 | Cheapest, good contrast and resolution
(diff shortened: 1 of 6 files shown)


martinberlin:
Ok I'm hitting here some strange validations from the auto checks CI:

 /opt/esp/tools/xtensa-esp-elf/esp-13.2.0_20240530/xtensa-esp-elf/bin/../lib/gcc/xtensa-esp-elf/13.2.0/../../../../xtensa-esp-elf/bin/ld: esp-idf/epdiy/libepdiy.a(epd_board_v7.c.obj): in function `epd_board_deinit':
/app/vroland/epdiy/src/board/epd_board_v7.c:148:(.text.epd_board_deinit+0x3): undefined reference to `epd_lcd_deinit'

I don't really see in VSCODE **epd_lcd_deinit** exits and it's referenced. It builds and flashes without any problem

vroland:
Not sure why the linker problem occurs, but could we split up adding the display and updating the examples into separate PRs? Maybe then the issue gets easier to debug as well.

martinberlin:
Sure later will do that.
@vroland what do you think is the best strategy to work out the mirroring X issue on this displays?
Is the best to make an **ifdef** or add it as a displays.c property?

vroland:
I'd add a displays.c property. But is it the horizontal or vertical direction for this display? Because the one where I had issues it was mirrored vertically. Do we have to account for both?

martinberlin:
> Because the one where I had issues it was mirrored vertically. Do we have to account for both?

Yes! I think it would be the best to add somehow both flags. This display needs only horizontal mirror but others will need the vertical mirroring also. I would add in displays.c properties a way to pass:

- X_MIRROR horizontal
- Y_MIRROR vertical   so we cover all possible cases

martinberlin:
Closing here since this is too old and the new display is already in #395

Issue #393, opened 2025-01-23
[Question] Detecting broken/smashed epaper displays by TPS65185 VOM kick-back voltage?


cyberman54:
I'm looking for a way to detect a smashed / broken epaper display with epdiy, just by software.
The idea is: The TPS65185 can perform a voltage measurement on the VCOM pin to determine the kick-back voltage of the epd panel.
Would it be expectable to detect a smashed epaper display by reading this value and compare it against a certain threshold value?
Has anyone tested this?

/cc @martinberlin

martinberlin:
They won't be other way than to test this. From the TPS65185 datasheet we can read:

**8.3.7.1 Kick-Back Voltage Measurement**

TPS65185x can perform a voltage measurement on the VCOM pin to determine the kick-back voltage of the panel. This allows in-system calibration of VCOM. To perform a kick-back voltage measurement, follow these
steps:
• Pull the WAKEUP pin and the PWRUP pin high to enable all output rails.
• Set the HiZ bit in the VCOM2 register. This puts the VCOM pin in a high-impedance state.
• Drive the panel with the Null waveform. Refer to E-Ink specification for detail.
• Set the ACQ bit in the VCOM2 register to 1. This starts the measurement routine.
• When the measurement is complete, the ACQC (Acquisition Complete) bit in the INT1 register is set and the
nINT pin is pulled low.
• The measurement result is stored in the VCOM[8:0] bits of the VCOM1 and VCOM2 register.
The measurement result is not automatically programmed into nonvolatile memory. Changing the power-up default is described in the following paragraph.

Questions: 

1. What is the null waveform?
2. Would be the same just to refresh white the display and measure it?
3. Would be another means to detect vandalism like a sudden change of temperature (TPS65185 has also a thermistor) or simply that the Firmware will hang since cannot update the display. It all depends on how bad you call it broken. Because it can be completely smashed or just have a break in the side and in that case I'm not sure if the Kick-Back Voltage measurement will mark any difference. 
Hopefully will do if it's broken but also we don't know because it really needs some experimentation.

vroland:
Sounds like an interesting thing to try! We haven't even implemented kick-back measurement for normal VCOM calibration yet though, so that would be a prerequisite. And we know that there is quite a variance in VCOM from display to display, but detecting a change for the same display sounds pretty plausible. As martin said, you'll need to try ;)

cyberman54:
Yes, I'm going to try it out together with Martin with an experiment.

[...]

PR #395, opened 2025-01-30
#393 Implement kickback voltage readings

martinberlin:
Please check Issue #393 for details
Additionally update the SSL certificate for www-image (Loremflickr.com) usually this are renewed every year and this will make the jpeg download example SSL handshake fail. 

Maybe Valentin what we can do is a bash or python script that just runs:

    openssl s_client -showcerts -connect www.loremflickr.com:443 </dev/null

And fetches last certificate in the chain to store it in https://github.com/vroland/epdiy/tree/main/examples/www-image/ssl_cert


martinberlin:
@vroland I must say that clang-format is a real loose of time. It really complains for stuff that alright like: 

```c
//tps65185.c:122:51: error: code should be clang-formatted [-Wclang-format-violations]
        int u16Value = (lsb | (msb << 8)) & 0x1ff;
 // Same here. I really do not get it!                                                 ^
       ESP_LOGI("Kickback raw readings", "value: %d TPS temperature: %d °C", u16Value, tps_read_thermistor(I2C_NUM_0));
```

src/board/epd_board_v7.c
```diff
@@ -248,6 +248,53 @@ static void epd_board_poweron(epd_ctrl_state_t* state) {
     }
 }
 
+static void epd_board_measure_vcom(epd_ctrl_state_t* state) {
```
vroland, line 251: Since this is the same as on v6, can we refactor it out into a reusable function so we don't duplicate it?

src/waveforms/epdiy_NULL.h
```diff
@@ -0,0 +1,28 @@
...
...00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00}},{{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00}},{{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00}},{{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00},{0x00,0x00,0x00,0x00}}};
```
vroland, line 2: Just for curiosity: Have you checked what happens if we put 0xFF everywhere instead of 0x00?

vroland (changes requested):
The kickback reading feature looks nice! I think the values we're getting are not quite the once we would expect from the labels, because the timing of the measurement is not the best. Ideally we would drive the waveform without pause until the measurement is complete, while currently we pause to do all the highlevel stuff. But that requires a lot more work and like this it can already be used for detecting broken displays and such, so I think we should bring it in :)

Before merging I'd prefer that the duplicate code (see comment) is refactored. Also I'd split out the www-image changes and the kickback voltage changes into separate PRs. On that note: Is there a specific reason for adding the whole jpegdec component now, when the only other change is the cert?

martinberlin:
> The kickback reading feature looks nice! I think the values we're getting are not quite the once we would expect from the labels, because the timing of the measurement is not the best. Ideally we would drive the waveform without pause until the measurement is complete, while currently we pause to do all the highlevel stuff. But that requires a lot more work and like this it can already be used for detecting broken displays and such, so I think we should bring it in :)

agree!

> Before merging I'd prefer that the duplicate code (see comment) is refactored. Also I'd split out the www-image changes and the kickback voltage changes into separate PRs. On that note: Is there a specific reason for adding the whole jpegdec component now, when the only other change is the cert?

Jpegdec falled behind since the author also added faster decode using assembly.
But maybe we could just take latest version from https://components.espressif.com/components/bitbank2/jpegdec/versions/1.6.2 ?
Otherwise I fully agree next time let’s keep the PRs only specific to one single topic…

src/waveforms/epdiy_NULL.h
martinberlin: No that could be an interesting take, but do you thing the kickback reading should return a similar value to the VCOM sticker?
That would be cool then VCOM could be detected for each display in all boards that already use the TI PMIC

src/board/epd_board_v7.c
martinberlin: Sorry Valentin but I'm not really going to this for two duplicate functions since needs refactoring many things and does NOT make any sense for a funcion that only 1% of the epdiy users will ever use.
So you can accept as it is, or you can propose a different refactoring, but I'm not willing to do it just for this. For more info: Doing this would imply doing a common_v6 (While there is already a common for the boards but from v2 to v5) so you need to really add more files. And it's not prepare to do that since it will imply that many variables + includes have to be moved there (Like #include "pca9555.h" #include "tps65185.h" etc)
And does not really make it just to avoid a single duplicate function.

martinberlin:
I will check over the weekend to see if I can get a better VCOM reading. But about the duplicate function please read my comment:  It really does not make sense to refactor all just because a duplicate function that hardly someone will use. It involves adding more files and changing the structure. Something it makes sense if we do a bigger refactoring but not because of this VCOM voltage reading.

src/waveforms/epdiy_NULL.h
martinberlin: With 0x00 gives me values around 500 and 540:
I (2417) vcom: raw value:50 temperature:19 C
I (2417) vcom: readings are of 500 mV.

With 0xFF gives me a bit lower values:

I (2413) vcom: readings are of 430 mV. Other reads 470 mV. It always changes a bit is not the same on every read but are lower of 500mV.

I think the readings are appearing way to early or the waveform needs to be faster to produce the right outcome. 
Now if you draw the dragon before or draw anything in the screen then the voltage changes. You should give it a try.

vroland: Ok, then it doesn't make a difference, good to know :) I think the kickback reading should be the thing on the label when done properly, to not have to set the calibrations manually in mass manufacturing. But this is definitely a start.

vroland:
Sure, I thought it would be a quick change to factor out that function, but if not it's fine. Only if we get more V7 variants we should think about refactoring that.

vroland:
About JPEGDEC: For me it works if I delete the 'components' directory entirely and just run `idf.py add-dependency "bitbank2/jpegdec^1.6.2"`. Then it creates a `main/idf_component.yaml` which makes the build system automatically download the component when I build. I think that's the most elegant solution :)

martinberlin:
> just run idf.py add-dependency "bitbank2/jpegdec^1.6.2"

Yes!  Deleted what is components and add only added idf_component.yaml 
So now it has less duplicated code. Only downside of doing this is that if you build on Arduino then you need to first install JPEGDEC as a library. But that is ok since is the Arduino IDE way to do stuff.
OK all green and it passes. Next PR will be to add the new v7 board without IO expander, that I've called "v7 raw" (Because it sounds a bit weird to call it "naked v7")

PR #413, opened 2025-04-12
Fix builds for versions with new lcd_hal_cal_pclk_freq

vroland:
There was a change in the parameters of this function in this [idf commit](https://github.com/espressif/esp-idf/commit/d39388fe4f4c5bfb0b52df9177307b1688f41016#diff-2df607d77e3f6e350bab8eb31cfd914500ae42744564e1640cec47006cc17a9c).
There are different builds with the same IDF minor version, some with, some without the commit. So we try to select the correct one by checking if the flag value is defined.


martinberlin:
Hello Valentin,
I can confirm now it builds and run correctly in IDF 5.4 and should also in 5.3 and main branches of the toolchain. 
Thanks for your quick fix

martinberlin (approved):
Looking good!

martinberlin:
@vroland please merge this when you find a second to make a click. It's working great and will avoid to have a bad experience at the start for many users

PR #416, opened 2025-05-04
Board/v7_raw adding new Board configuration

martinberlin:
This new open source board Hardware is located here:
https://github.com/vroland/epdiy-hardware/tree/main/epdiy-v7-raw

Adding support in the Firmware side.

@vroland I've just a doubt that I don't really know why it happens. Using the following configuration in the dragon example:

```c
    epd_init(&epd_board_v7_raw, &ED052TC4, EPD_LUT_64K);
    epd_set_vcom(1560);
    hl = epd_hl_init(EPD_BUILTIN_WAVEFORM);
```

The dragon is correctly drawn but the first 20 pixels or so are completely white. Since you have one of the first releases of this board maybe you can try it out?

Thanks in advance


martinberlin:
@vroland please review this when you have some time. 
After this is merged would like to make an additional PR also with v7-usb "as a device" examples
https://github.com/vroland/epdiy/tree/v7/usb-device/examples/v7-usb-device
Video demo: https://x.com/martinfasani/status/1921095661866438990

Reading images from SD-Card and making a fast offline gallery

martinberlin:
Oh yes sorry I’ve to remove that it was an internal test. By the way only thing that bothers to me adding this board is that this panel has a X mirror, so drawing fonts won’t be nice. Is there any way we could add X mirroring option?
Like width() -x.
Would be nice if we don’t need to add a conditional for every pixel drawn like to do it in a lower level?
Does that make sense?

martinberlin:
Shall we merge it as is and then plan the X mirroring in a further PR ?

PR #434, opened 2025-07-23
Fixes #433 adding mirror x

### LINKED ISSUE #433: Add MIRROR_X before row output
Currently some displays like ED103TC2 have the problem that the output is mirrored.
This could be solved adding a:  epd_width() -x  in the drawpixel function but is IMHO not the right way to do it. Because if you use icons or images with the image script that are directly outputted in the framebuffer skipping any draw fuction, then you will still have the mirroring issue. 

@vroland I will make a first attempt to fix this in case I fail maybe you can point me in the right direction.

I asked CoPilot and there is a mirror function for some displays, but I don't think it exists for our particular use-case:

> Yes, the ESP-IDF LCD module does provide a way to mirror the display output along the X axis (as well as the Y axis) for many supported LCD panels.

How it works:
The API function esp_lcd_panel_mirror(panel, mirror_x, mirror_y) can be used to enable or disable mirroring about the X and/or Y axes.
This function is available for multiple drivers, including RGB, ST7789, SSD1306, and NT35510 panels, as shown in their respective source files.

martinberlin:
It’s for me the easiest here to add it in rotation without making an extra function.
Display type ED097TC2 is something that affects only old I2S version and this seems to far to appear only in modern displays (16 bit - S3 version)


jonasschnelli:
Tested ACK 16f0848edcf204e0e5819ab445ef71d190e03c5c

martinberlin:
BUG discovered. Please do not merge this.
Even using standard rotation: epd_set_rotation(EPD_ROT_LANDSCAPE);

When you use the MIRROR_X displayType is disturbing the partial update. It starts to work only after 50% of the display. And that's not good. In order to test it you can just simulate a progressBar and call it in a loop:

```c
void progressBar(long processed, long total)
{
  float progress = 0.0f;
  progress = (float)total / processed;
  float perc =(float)epd_width() / progress;
  EpdRect rectangle = {
    .x = 0,
    .y = 0,
    .width = (int)perc,
    .height = progressBarHeight
  };
  epd_fill_rect(rectangle, 0, fb);
  epd_hl_update_area(&hl, MODE_DU, 25, rectangle);
}
```

martinberlin:
Closing here let us try #436

PR #436, opened 2025-08-01
add optional horizontal mirroring for framebuffers

jonasschnelli:
Some panels (ED052TC4, ED103TC2) have horizontally mirrored layouts requiring data transformation. ESP functions like `esp_lcd_panel_mirror()` can't be used due to our HAL-level LCD usage. Implementing mirroring in the hardware layer proved tricky and caused timing issues in experiments.

This implementation mirrors the entire framebuffer at the software layer rather than modifying all drawing functions or hardware timing.

- Add mirror_x flag to EpdiyHighlevelState
- Mirror both front and back framebuffers before display update
- Restore original state after hardware processing
- Optimized scalar implementation with 4x loop unrolling

Performance (**acceptable IMO**): ~80ms total for 1280x720 on ESP32-S3 (4 operations × 20ms)


jonasschnelli:
Fixes #433 alternative to #434

martinberlin:
Great then you can merge it directly or I can do it end of this week. Thanks for your review and thanks @jonasschnelli for the new feature

Issue #451, opened 2025-12-17
Add different UPSEQ voltage initialisation sequence for new displays such as ED103MC2


martinberlin:
This comes from a display that @jonasschnelli sent me. 
It turns out that the new displays of Eink have a special up-sequence which we can also set and is described in page 19 of https://www.ti.com/lit/ds/symlink/tps65185.pdf

The UPSEQ0 and 1 sequence needs to be specified just after WAKE UP and before PWRUP.

This was not a trivial update and took me 3 weeks of research. I had to buy a 10.3 Waveshare controller and a Logic analyser to sniff the I2C signals and use the same UPSEQ pattern. Without using it you simply get: VNEG undervoltage and the rails do not power up.

PR #452, opened 2025-12-17
#451 Implement UPSEQ

martinberlin:
Implements a solution for #451


vroland:
Hey, nice work on finding this out! I'm wondering though, why implement it as a new board type? Does it actually require different hardware? Since it may also be useful to enable with normal v7.

vroland:
Did you check if that power-on sequence impairs any of the other displays? Since it's mentioned in the docs, maybe it should just be the default?

martinberlin:
Great then I can refactor this and do it for v6, v7. Although v6 didn’t had 16 bit display support maybe there will be some 8 bit panels requiring this up sequence.
Good idea.

PR #460, opened 2026-01-25
fix: Handle pre-existing I2C driver gracefully in epd_board_v7

Bwooce:
## Problem

When using epdiy on boards with other I2C peripherals (touch controllers, RTCs, battery gauges), users often initialize Arduino's `Wire.begin()` before calling `epd_init()`. This causes `i2c_driver_install()` to fail with `ESP_ERR_INVALID_STATE` because the driver is already installed.

Similarly, `gpio_install_isr_service()` can fail if other code has already installed the ISR service.

## Solution

Check return values and accept `ESP_ERR_INVALID_STATE` as success, allowing epdiy to reuse the existing I2C driver and ISR service instead of failing.

## Changes

- `epd_board_v7.c`: Check `i2c_param_config()`, `i2c_driver_install()`, and `gpio_install_isr_service()` return values
- Accept `ESP_ERR_INVALID_STATE` as non-fatal (driver/service already exists)

## Testing

Tested on LilyGo T5 S3 E-Paper Pro sharing I2C bus with:
- GT911 touch controller
- PCF8563 RTC
- BQ25896/BQ27220 battery management

All peripherals coexist on the same I2C bus (SDA=39, SCL=40).

### DIFF
--- a/src/board/epd_board_v7.c
+++ b/src/board/epd_board_v7.c
@@ -80,6 +80,8 @@ static int vcom = 1600;
 static epd_config_register_t config_reg;
 
 static bool interrupt_done = false;
+static bool i2c_driver_installed_by_us = false;
+static bool isr_service_installed_by_us = false;
 
 static void IRAM_ATTR interrupt_handler(void* arg) {
     interrupt_done = true;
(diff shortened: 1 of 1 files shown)


martinberlin:
Please remove the unnecessary commit that brought this in:

#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
     PERIPH_RCC_ATOMIC()
     
The changes I [asked for here](https://github.com/vroland/epdiy/pull/460#issuecomment-3803828852) are already merged in main.
 After removing that commit I mentioned, please switch to main branch and do a  **pull origin main**, merge main back here, and then we will run all the test on this PR. Thanks for your collaboration

martinberlin:
Sorry this won't be merged as is. Please make a new PR based on latest main and copy from here only the necessary.

PR #461, opened 2026-01-25
fix: Restore IDF 4.x compatibility in lcd_driver.c

Bwooce:
## Problem

The current `lcd_driver.c` uses IDF 5.x APIs (`driver/rmt_tx.h`, etc.) which are not available in IDF 4.x. This breaks compatibility with Arduino ESP32 framework which still uses IDF 4.x.

## Solution

Add `ESP_IDF_VERSION` conditionals to support both IDF 4.x and IDF 5.x, similar to other parts of the codebase.

## Changes

- Add version conditionals for IDF 5.x vs 4.x includes (`driver/rmt_tx.h` vs `driver/rmt.h`)
- Add IDF version conditional for `rmt_ll_set_group_clock_src()` signature difference
- Wrap `RMTMEM` extern declaration for IDF 5.x only
- Add fallback defines for `PERIPH_LCD_CAM_MODULE` and `PERIPH_RMT_MODULE` if not defined
- Include `idf-4-backports.h` for required backported functions on IDF 4.x

## Testing

Tested with PlatformIO + Arduino ESP32 framework (espressif32 platform) on LilyGo T5 S3 E-Paper Pro.

## Note

These changes are based on the IDF compatibility work from LilyGo's T5S3-4.7-e-paper-PRO SDK.

### DIFF
--- a/src/output_lcd/lcd_driver.c
+++ b/src/output_lcd/lcd_driver.c
@@ -66,14 +66,6 @@ gpio_hal_context_t hal = { .dev = GPIO_HAL_GET_HW(GPIO_PORT_0) };
 #define LCD_PERIPH_SIGNALS lcd_periph_rgb_signals
 #endif
 
-// Use named peripheral module defines where available, struct-based as fallback
-#ifndef PERIPH_LCD_CAM_MODULE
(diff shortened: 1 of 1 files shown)


vroland (approved):
Sure, I think this is small enough to not be too much of a hassle, just note that testing for 4.x is not something we'll actively do necessarily.

PR #463, opened 2026-01-26
Add ESP-IDF 6.0 support with dual 5.x/6.x compatibility

### LINKED ISSUE #450: Make it possible to compile in ESP-IDF version 6.1
Recently got a new Mac mini M4 and just to try it, downloaded latest **main** branch from ESP-IDF, this is result of trying to build Epdiy.
This happened in every major release so it does not surprise me at all. Maybe important also for @jonasschnelli and @mickeprag 

1. It seems **soc/lcd_periph.h** is not anymore there
epdiy/src/output_lcd/lcd_driver.c:14:10: fatal error: soc/lcd_periph.h: No such file or directory
   14 | #include <soc/lcd_periph.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.

2. epdiy/src/board/epd_board_common.c:1:10: fatal error: **driver/adc.h**: No such file or directory
    1 | #include "driver/adc.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.

So far this 2 things seem to have moved somewhere else. Let's see what comes additionally once that is fixed.

Bwooce:
## Summary

Adds ESP-IDF 6.0 support while maintaining full backward compatibility with IDF 5.x (tested against 5.5). All changes use `#if ESP_IDF_VERSION` guards so both major versions compile cleanly for both ESP32 and ESP32-S3 targets.

Closes #450

### Build system (`CMakeLists.txt`)
- Added IDF 6.x branch for component registration: IDF 6 splits legacy `driver` into sub-components (`esp_hal_dma`, `esp_driver_rmt`, `esp_driver_gpio`), so these are now listed as explicit REQUIRES for `IDF_VERSION_MAJOR > 5`
- Existing IDF 4.x and 5.x branches are untouched

### LCD driver — ESP32-S3 (`src/output_lcd/lcd_driver.c`)
- **LCD peripheral signals**: struct renamed from `lcd_periph_signals.panels[0].member` / `lcd_periph_rgb_signals.panels[0].member` to array `soc_lcd_rgb_signals[0].member` in IDF 6. Replaced `LCD_PERIPH_SIGNALS` macro with `LCD_PERIPH_SIG(member)` to abstract the access pattern across all three naming conventions
- **LCD signal struct**: forward-declared `soc_lcd_rgb_signal_desc_t` and extern `soc_lcd_rgb_signals` directly instead of including `hal/lcd_periph.h`, which in IDF 6 transitively pulls in `hal/i2s_ll.h` causing conflicts
- **GDMA API**: `gdma_new_channel()` replaced by `gdma_new_ahb_channel()` in IDF 6 — takes 3 args (config, tx_chan, rx_chan) and `gdma_channel_alloc_config_t` no longer has a `direction` member
- **RMT peripheral clock**: `periph_module_enable(PERIPH_RMT_MODULE)` removed in IDF 6 (`PERIPH_RMT_MODULE` no longer exists in `shared_periph_module_t`). Replaced with `rmt_ll_enable_bus_clock()` + `rmt_ll_reset_register()` inside `PERIPH_RCC_ATOMIC()` blocks
- **RMT LL rename**: `rmt_ll_enable_periph_clock()` renamed to `rmt_ll_enable_group_clock()` in IDF 6
- **RMT types**: `rmt_item32_t` removed with legacy RMT driver in IDF 6 — defined locally with identical layout
- **RMT signals struct**: `rmt_periph_signals.groups[0]` renamed to array `soc_rmt_signals[0]` in IDF 6
- **RMT channel count**: `SOC_RMT_CHANNELS_PER_GROUP` removed — replaced with `RMT_LL_CHANS_PER_INST` from `hal/rmt_ll.h`
- **RMT periph header**: moved from `soc/rmt_periph.h` to `hal/rmt_periph.h` in IDF 6
- **LCD LL interrupt macro**: IDF 5.5+ added a wrapper macro around `lcd_ll_enable_interrupt()` that references `__DECLARE_RCC_ATOMIC_ENV`, requiring the caller to be inside a `PERIPH_RCC_ATOMIC()` block. Guard lowered to `>= 5.5.0` (not just 6.0)

### RMT pulse driver — ESP32 (`src/output_i2s/rmt_pulse.c`)
- **Legacy RMT driver removed**: `driver/rmt.h` (`rmt_config()`, `rmt_set_tx_intr_en()`, `rmt_config_t`) completely removed in IDF 6. Replaced init with direct register writes matching what `rmt_config()` did internally — the actual pulse generation already used direct RMTMEM access
- **RMT types**: locally defined `rmt_item32_t` and `rmt_block_mem_t` for IDF 6 (same approach as lcd_driver.c)
- **Peripheral clock**: same `PERIPH_RCC_ATOMIC()` + LL function pattern as lcd_driver.c
- **GPIO routing**: `gpio_matrix_out()` replaced with `esp_rom_gpio_connect_out_signal()` via `esp_rom_gpio.h`

### I2S data bus — ESP32 (`src/output_i2s/i2s_data_bus.c`)
- **Peripheral clock**: `periph_module_enable(PERIPH_I2S1_MODULE)` removed in IDF 6 — replaced with `i2s_ll_enable_bus_clock(1, true)` + `i2s_ll_reset_register(1)` inside `PERIPH_RCC_ATOMIC()`. Same for disable in `i2s_bus_deinit()`
- **GPIO signal macros**: `I2S1O_DATA_OUT0_IDX` / `I2S1O_WS_OUT_IDX` lost their transitive include path in IDF 6 — added explicit `#include "soc/gpio_sig_map.h"`
- **GPIO routing**: `gpio_matrix_out()` replaced with `esp_rom_gpio_connect_out_signal()` for IDF 6

### ADC temperature sensing (`src/board/epd_board_common.c`)
- **Legacy ADC API removed**: `adc1_get_raw()`, `esp_adc_cal_characterize()`, `esp_adc_cal_raw_to_voltage()` all removed in IDF 6
- Reimplemented `epd_board_temperature_init_v2()` and `epd_board_ambient_temperature_v2()` using the oneshot ADC driver (`adc_oneshot_new_unit`, `adc_oneshot_config_channel`, `adc_oneshot_read`) + new calibration API (`adc_cali_create_scheme_line_fitting`, `adc_cali_raw_to_voltage`)

### Cache header path (`src/output_lcd/render_lcd.c`)
- `rom/cache.h` may be relocated in IDF 6 — added `__has_include` fallback to `esp32s3/rom/cache.h`

### SPIRAM config (`src/highlevel.c`)
- IDF 6 unifies `CONFIG_ESP32_SPIRAM_SUPPORT` / `CONFIG_ESP32S3_SPIRAM_SUPPORT` under `CONFIG_SPIRAM` — added it as an additional check

### Demo build (`examples/demo/main/CMakeLists.txt`)
- GCC 15 (shipped with IDF 6) treats `-Wbidi-chars=` as an error for Unicode bidirectional control characters in the generated font header files. Added `-Wno-error=bidi-chars=` compile option

### CI (`.github/workflows/main.yml`)
- Added `release-v6.0` to the build matrix for both ESP32 and ESP32-S3 (switch to `v6.0` once the stable Docker tag exists)
- Added IDF 6 extended examples (mirrors the v5.4 set)
- Set `continue-on-error: true` for IDF 6 while it is pre-release
- IDF 6 kconfgen crashes on large sdkconfig.defaults from IDF 5 (upstream bug: `choice.selection` is None). CI uses minimal sdkconfig.defaults for IDF 6 builds as a workaround
- Added comment noting Arduino-ESP32 IDF 6 support is pending upstream

## CI results

| Target | IDF 5.2 | IDF 5.3 | IDF 5.4 | IDF 6.0 |
|--------|---------|---------|---------|---------|
| ESP32-S3 demo | PASS | PASS | PASS | PASS |
| ESP32 demo | PASS | PASS | PASS | PASS |
| screen_diag | — | — | PASS | PASS |
| dragon | — | — | PASS | PASS |
| grayscale_test | — | — | PASS | PASS |
| calibration_helper | — | — | PASS | PASS |
| www-image | — | — | PASS | FAIL (*) |
| Arduino weather | PASS (v5.3.2) | | | |

(*) `www-image` fails on IDF 6 due to GCC 15 warnings-as-errors in the third-party `bitbank2/jpegdec` managed component (`-Werror=pointer-to-int-cast`, `-Werror=shift-negative-value`), not epdiy code. Requires an upstream fix to jpegdec.

## Test plan

- [x] CI passes for IDF 5.2, 5.3, 5.4 (all green)
- [x] CI passes for IDF 6.0 ESP32 and ESP32-S3 demo builds
- [x] CI passes for IDF 6.0 extended examples (except www-image — third-party dependency)
- [ ] Verify demo runs correctly on physical ESP32-S3 hardware with IDF 6.0
- [ ] Verify demo runs correctly on physical ESP32 hardware with IDF 6.0
- [ ] Once IDF 6.0 reaches stable release, switch `release-v6.0` to `v6.0` and remove `continue-on-error`

### DIFF
--- a/src/board/epd_board_common.c
+++ b/src/board/epd_board_common.c
@@ -1,7 +1,84 @@
-#include "driver/adc.h"
 #include "epd_board.h"
-#include "esp_adc_cal.h"
 #include "esp_log.h"
+
+#include <esp_idf_version.h>
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
+// IDF 6.x: legacy ADC driver removed, use oneshot driver + calibration
+#include "esp_adc/adc_oneshot.h"
+#include "esp_adc/adc_cali.h"
+#include "esp_adc/adc_cali_scheme.h"
+
+static adc_oneshot_unit_handle_t adc_handle = NULL;
+static adc_cali_handle_t adc_cali_handle = NULL;
+static const adc_channel_t channel = ADC_CHANNEL_7;
+
+#define NUMBER_OF_SAMPLES 100
+
+void epd_board_temperature_init_v2() {
+    // Initialize ADC oneshot unit
+    adc_oneshot_unit_init_cfg_t unit_cfg = {
+        .unit_id = ADC_UNIT_1,
+    };
+    ESP_ERROR_CHECK(adc_oneshot_new_unit(&unit_cfg, &adc_handle));
+
+    // Configure channel
+    adc_oneshot_chan_cfg_t chan_cfg = {
+        .atten = ADC_ATTEN_DB_6,
+        .bitwidth = ADC_BITWIDTH_12,
+    };
+    ESP_ERROR_CHECK(adc_oneshot_config_channel(adc_handle, channel, &chan_cfg));
+
+    // Initialize calibration
+#if ADC_CALI_SCHEME_CURVE_FITTING_SUPPORTED
+    adc_cali_curve_fitting_config_t cali_cfg = {
+        .unit_id = ADC_UNIT_1,
+        .atten = ADC_ATTEN_DB_6,
+        .bitwidth = ADC_BITWIDTH_12,
+    };
+    esp_err_t ret = adc_cali_create_scheme_curve_fitting(&cali_cfg, &adc_cali_handle);
+#elif ADC_CALI_SCHEME_LINE_FITTING_SUPPORTED
+    adc_cali_line_fitting_config_t cali_cfg = {
+        .unit_id = ADC_UNIT_1,
+        .atten = ADC_ATTEN_DB_6,
+        .bitwidth = ADC_BITWIDTH_12,
+    };
+    esp_err_t ret = adc_cali_create_scheme_line_fitting(&cali_cfg, &adc_cali_handle);
+#else
+    esp_err_t ret = ESP_ERR_NOT_SUPPORTED;
+#endif
+    if (ret == ESP_OK) {
+        ESP_LOGI("epd_temperature", "ADC calibration scheme initialized");
+    } else {
+        ESP_LOGW("epd_temperature", "ADC calibration not available (err=%d), raw values used", ret);
+    }
+}
+
+float epd_board_ambient_temperature_v2() {
+    int raw = 0;
+    uint32_t value = 0;
+    for (int i = 0; i < NUMBER_OF_SAMPLES; i++) {
+        if (adc_oneshot_read(adc_handle, channel, &raw) == ESP_OK) {
+            value += raw;
+        }
+    }
+    value /= NUMBER_OF_SAMPLES;
+    // voltage in mV
+    int voltage_mv = 0;
+    if (adc_cali_handle != NULL) {
+        adc_cali_raw_to_voltage(adc_cali_handle, (int)value, &voltage_mv);
+    } else {
+        // Fallback: rough estimate without calibration
+        // 12-bit ADC with ~6dB attenuation gives roughly 0-2200mV range
+        voltage_mv = (int)(value * 2200 / 4095);
+    }
+    return ((float)voltage_mv - 500.0f) / 10.0f;
+}
+
+#else
+// IDF 4.x / 5.x: use legacy ADC driver
+#include "driver/adc.h"
+#include "esp_adc_cal.h"
 
 static const adc1_channel_t channel = ADC1_CHANNEL_7;
 static esp_adc_cal_characteristics_t adc_chars;
@@ -32,3 +109,4 @@ float epd_board_ambient_temperature_v2() {
     float voltage = esp_adc_cal_raw_to_voltage(value, &adc_chars);
     return (voltage - 500.0) / 10.0;
 }
+#endif
--- a/src/output_i2s/rmt_pulse.c
+++ b/src/output_i2s/rmt_pulse.c
@@ -78,7 +78,7 @@ void rmt_pulse_init(gpio_num_t pin) {
     RMT.conf_ch[RMT_PULSE_CHANNEL].conf0.carrier_out_lv = 0;
     RMT.conf_ch[RMT_PULSE_CHANNEL].conf1.tx_conti_mode = 0;  // loop_en = false
     RMT.conf_ch[RMT_PULSE_CHANNEL].conf1.idle_out_en = 1;
-    RMT.conf_ch[RMT_PULSE_CHANNEL].conf1.idle_out_lv = 0;    // idle level low
+    RMT.conf_ch[RMT_PULSE_CHANNEL].conf1.idle_out_lv = 0;  // idle level low
 
     // Connect GPIO to RMT channel
     gpio_set_direction(pin, GPIO_MODE_OUTPUT);
--- a/src/output_i2s/rmt_pulse.c
+++ b/src/output_i2s/rmt_pulse.c
@@ -3,17 +3,15 @@
 
 #ifdef RENDER_METHOD_I2S
 
-#include "driver/rmt.h"
+#include <esp_idf_version.h>
+
 #include "rmt_pulse.h"
 
 #include "soc/rmt_struct.h"
 
 static intr_handle_t gRMT_intr_handle = NULL;
 
-// the RMT channel configuration object
-static rmt_config_t row_rmt_config;
-
-// keep track of wether the current pulse is ongoing
+// keep track of whether the current pulse is ongoing
 volatile bool rmt_tx_done = true;
 
 /**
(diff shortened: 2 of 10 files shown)


martinberlin:
@Bwooce please fix as soon as possible the failed automatic checks if you want to get this merged

Bwooce:
All checks are now passing. The one remaining failure is `www-image` on IDF 6.0, which is caused by GCC 15 warnings-as-errors in the third-party `bitbank2/jpegdec` managed component (shift-negative-value, pointer-to-int-cast), not epdiy code. I've submitted a fix upstream (https://github.com/bitbank2/JPEGDEC/pull/113), waiting on their review.

The IDF 6.0 jobs have `continue-on-error: true` so the overall workflow reports success, but the individual check still shows as failed.

martinberlin:
@Bwooce what is the status here?
Shall I write Larry Bank and ask about an update in JPEGDEC?

I will try to add your PR about I2C next weekend. Sorry but it was stale, and not working properly, in the tests I run.

Bwooce:
I'll go chase the missing upstream PR Martin, it's not a very active repo. I could also try and work around it, it's literally due to a "warnings now treated as errors" change in IDF 6 which uses GCC 15 so the PR is nice-to-have functionally but working around it long-term would be incorrect.

martinberlin:
No problem. I will try to test your update. I've a v7 board to share with you if you would like to try it, just need to pay the shipping.

src/board/epd_board_common.c
```diff
@@ -1,8 +1,85 @@
-#include "driver/adc.h"
 #include "epd_board.h"
-#include "esp_adc_cal.h"
 #include "esp_log.h"
 
+#include <esp_idf_version.h>
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
```
vroland, line 6: I think this may not need to be conditional, since the oneshot adc driver was introduced in IDF V5, and we don't support IDF 4.x anymore. Then we remove the old implementation. Could also be moved into a separate PR.

src/output_i2s/rmt_pulse.c
```diff
@@ -26,12 +24,116 @@ static void IRAM_ATTR rmt_interrupt_handler(void* arg) {
...
+
+// Channel index used for RMT pulse generation
+#define RMT_PULSE_CHANNEL 1
+#define RMT_MEM_OWNER_TX 1
+
+#include <esp_private/periph_ctrl.h>
+#include <driver/gpio.h>
+#include "esp_rom_gpio.h"
+
+void rmt_pulse_init(gpio_num_t pin) {
```
vroland, line 66: Same here, I don't think we gain much by keeping both implementations. We can just migrate to the new implementation. Also maybe in a separate PR?

src/output_lcd/lcd_driver.c
```diff
@@ -79,17 +115,48 @@ static inline int max(int x, int y) {
...
 // The extern line is declared in esp-idf/components/driver/deprecated/rmt_legacy.c. It has access
 // to RMTMEM through the rmt_private.h header which we can't access outside the sdk. Declare our own
-// extern here to properly use the RMTMEM smybol defined in
+// extern here to properly use the RMTMEM symbol defined in
 // components/soc/[target]/ld/[target].peripherals.ld Also typedef the new rmt_mem_t struct to the
 // old rmt_block_mem_t struct. Same data fields, different names
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
+// IDF 6.x: rmt_mem_t and rmt_item32_t removed with legacy driver.
+// Define compatible types for direct RMT memory access.
+#include <soc/rmt_reg.h>
```
vroland, line 134: This would be included in the rmt migration, then it doesn't need to be duplicated in I2S and LCD driver.

vroland (commented):
Hi, sorry to be late to the party, just a suggestion if it doesn't burn too many tokens: We could separate the ADC and RMT migrations into separate PRs, since those should be backwards compatible. The ADC I can test while travelling, the RMT I can't test right now because I don't have an older board with me.
Then we can rebase the 6.0 branch, which should become much smaller. At that point, if the JPEGDEC dependency is not updated, I'd be fine with moving the www-image example into a separate repo that only builds with 5.x or hacks, while all the remaining examples should build cleanly with 6.0.

martinberlin:
@vroland in the top of that further upgrades of IDF won't let you use the old I2C legacy driver. This happened already with me on a C5 board (Yes C5 works also to drive parallel using the new [Parlio driver](https://github.com/bitbank2/FastEPD/tree/esp32c5))

https://github.com/martinberlin/i2c_bus this is the super simple component I built so you can just use one I2C bus for many different I2C peripherals without running into the warning "I2C is already initialised" this would be also something cool to implement into epdiy. As many people uses it for home automation, connecting other i2c sensors to the board.

vroland:
If I understand your component correctly, it would still require people to use your component if they want to re-use the I2C bus? If that's the case, wouldn't it make more sense to migrate to the new I2C driver and provide a way to optionally initialize epdiy with an existing bus? Then we don't introduce another custom way of doing things.

martinberlin:
> If I understand your component correctly, it would still require people to use your component if they want to re-use the I2C bus? If that's the case, wouldn't it make more sense to migrate to the new I2C driver and provide a way to optionally initialize epdiy with an existing bus? Then we don't introduce another custom way of doing things.

Yeah sure what I was saying is that it would be nice to use something like this inside epdiy itself. So then we have only one file to modify where I2C is handled making it easy to update it without searching in what other parts of the code is instantiated (PCA9535, TPS65185) 
Just an idea that I though can bring a more cohesive and maintainable code

vroland:
Superseeded by #477

PR #477, opened 2026-05-11
IDF-6.0 support ported from #463

vroland:
This PR adds the remaining ESP-IDF 6.0 support work from #463 on top of current main, after the ADC and RMT migrations from that original PR were split out and addressed separately. It updates CI to build against the stable v6.0 image, adds the remaining IDF 6 build requirements and SDK config workaround, and adapts the ESP32 I2S and ESP32-S3 LCD paths for IDF 6 API/header changes while reusing the existing rmt_compat abstraction rather than carrying over the older duplicated RMT compatibility code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,7 @@ jobs:
           - v5.2
           - v5.3
           - v5.4
+          - v6.0
         target:
           - esp32
           - esp32s3
@@ -44,6 +45,18 @@ jobs:
             example: calibration_helper
           - version: v5.4
             example: test
+          - version: v6.0
+            example: screen_diag
+          - version: v6.0
+            example: dragon
+          - version: v6.0
+            example: grayscale_test
+          - version: v6.0
+            example: www-image
+          - version: v6.0
+            example: calibration_helper
+          - version: v6.0
+            example: test
 
     continue-on-error: ${{ matrix.version == 'latest' }}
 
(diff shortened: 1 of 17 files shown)


geiseri:
I have been using this on the lilygo 4.7 esp32 version (not esp32s) and I keep getting stuck at: [rmt_pulse.c#L51)](https://github.com/vroland/epdiy/blob/bdb85cce9fe215db73b9901ba7ff43565bd16e09/src/output_i2s/rmt_pulse.c#L51)
If I make this change:

```
diff --git a/src/output_i2s/rmt_pulse.c b/src/output_i2s/rmt_pulse.c
index 5c0d49b..dcf726c 100644
--- a/src/output_i2s/rmt_pulse.c
+++ b/src/output_i2s/rmt_pulse.c
@@ -12,8 +12,8 @@ static volatile bool rmt_tx_done = true;
 
 static void IRAM_ATTR rmt_interrupt_handler(void* arg) {
     (void)arg;
-    rmt_tx_done = true;
     rmt_compat_clear_interrupts();
+    rmt_tx_done = true;
 }
 
 void rmt_pulse_init(gpio_num_t pin) {
```

It doesn't lock up.  I think it might be related to activity on the mcu since it happens most often during wifi station scans and transfers that last more than a second or two.   That being said I am confused since all the clear does is sets a value.

martinberlin:
@geiseri please raise this concerns to the Lyligo team. 
epdiy has actually their own boards and we are not meant to be the Lilygo support team. So the Lilygo matters should explicitly take place over their own repositories and not here. Thanks for your understanding

vroland:
@geiseri Good catch, this is also a possible race condition on our boards, depending on the application. I've also fixed a few similar ones.

vroland:
I still have a strange issue with this where the output is vertically stretched on with a v7 board and a ED060XC5 display, so that only 2/3 of the picture is visible. Works fine on 5.4. not sure where else to dig, maybe I'll have to check with my scope when I have the time.

vroland:
@martinberlin Should be ready now, tested with a V7 board and V6 board. Maybe give it a quick review if you have the time :)

PR #479, opened 2026-05-21
ci: update checkout action version, simplify checkout, prevent duplicate runs

pfeerick:
What started simply as a bump to the checkout action version has evolved into 
- bump actions/checkout to v6 - [per GitHub node.js 24 migration](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)
- arduino-esp32 checkout step didn't need to install git, just use the checkout action again - it is faster and does the same job!
- trigger scope updated so updates to PRs don't result in double run of CI workflow
- pin espressif/esp-idf-ci-action to v1 - [per the recommendation in the action docs](https://github.com/espressif/esp-idf-ci-action/tree/v1)
- install clang-format directly rather than via esp-idf. for a straightforward format/lint check, we don't need the esp-format version as we aren't interested in code generation or target support... only C/C++ syntax. 

I'm pretty confident of the first four changes, but I understand if the last one isn't wanted, and can remove that if desired. However, one thing to keep in mind is the difference of approx 20s for the format-check action to run, vs 90 - 120s depending on the runner workloads (not to mention the ~3.59 GB image download every run).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,14 +9,11 @@ on:
 jobs:
   format-check:
     runs-on: ubuntu-latest
-    container:
-      image: "espressif/idf:v5.4"
     steps:
       - uses: actions/checkout@v6
+      - name: Install clang-format
+        run: sudo apt-get update -qq && sudo apt-get install -y -qq clang-format-18
       - run: |
-          . $IDF_PATH/export.sh
-          idf_tools.py install esp-clang
-          . $IDF_PATH/export.sh
           which clang-format
           make format-check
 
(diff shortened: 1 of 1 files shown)


martinberlin:
Hello @pfeerick 
All looks good for me. Assigned it to Valentin so he can take a last look but it should be merged soon since all Checks pass. Thanks for your interest and collaboration in this repository

vroland (approved):
Looks good, thanks! There used to be differences between the clang-format version available in the runner image and esp-idf, but apprantly not anymore.

PR #480, opened 2026-05-25
Add a dual-polarity monochrome mode for 1bpp (8ppB) buffers

zvirja:
I am trying to re-implement existing LilyGo T5 4.7 monochrome driver to use the latest version of this library and I faced an issue that we don't have an update mode for monochrome screen which drives both black and white pixels at the same time. We only have mode to either render from white or from black (so we drive the opposite).

Unfortunately, on my real LilyGo that produces quite low contrast and ghosting. The only reliable solution is to render many times (like 20 cycles) where I drive all pixels all the time (both white and black). 

With help of AI I generated and verified that a new lookup table works. Now if we don't specify that previous state was black/white, then we drive all the pixels. It allows to have clean and nice implementation in my driver.

Let me know if you are okay with this. I don't see any issues and to it's a reasonable behavior.


zvirja:
@martinberlin I am not sure if you are the right one, but I saw you active around. Could you please take a look at this one and confirm if there are any chances we could have this one merged? It's a quite trivial change and it unblock me to have a better driver for lilygo. 

Thanks!

martinberlin:
This "as is" won't be merged. It lacks context, is not clear if this update affects only this Lilygo or all eink driving for all models (Which I think it does). 
I don't mind if it's AI generated or not but should be human tested with different boards and displays. Assigning this to @vroland

zvirja:
@martinberlin  I would be happy to provide context, I tried my best in the description.

The issue is that for LilyGo the existing Wave function when we drive only black or white is not working great - it leads to low contrast and ghosting.

Currently for 8PPB we require to either come from totally black or totally white screen. Then you drive either only white or only black. In this PR I am filling the gap by adding mode where we drive both black and white at the same time. I would say that this mode must have been there from the beginning and I don't see any good reason for not having it.

I tested in intensively on the LilyGo 4.7 and could verify that it's the only approach that gives a good contrast.

I would be more than happy to provide more info - just clarify what you need.

P.S. The change is quite trivial, so we could just disregard the AI part entirely.

vroland:
I'm not opposed to adding it, but we should document the risk of burn-in. If we drive the pixels hard like this all the time, it may lead to pixels getting "stuck" over time. So the display might degrade much quicker with this approach. I think we can add this, but maybe log a warning when this mode is selected that you use it that it may degrade your display faster and you should use it at your own risk?

martinberlin:
> I think we can add this, but maybe log a warning when this mode is selected that you use it that it may degrade your display faster and you should use it at your own risk?

Fully agree with this. It would be cool also to add an 1BPP example since currently all epdiy examples are done for 16 grays (4 bits per pixel) 
Then at least it will be also better documented and with an example @zvirja 
Could be just a very simple example that draws a 

**"Hello this is a 1 BPP monochrome example"** and any random things you would like to add.

PR #481, opened 2026-06-06
Bound V7 power transitions and expose checked failures

rktdno:
A V7 display whose power-good signal never arrives can hang its calling task. An outer retry counter alone does not bound this: the PCA9555 and TPS65185 helpers previously passed an infinite timeout to the I2C driver. Returning from the void power-on callback also gives an application no way to skip drawing after failure.

This change adds monotonic 500 ms deadlines per power-good phase, 10 ms polling with at least one FreeRTOS tick, and 50 ms timeouts for PCA/TPS transfers. The named constants address the tick-rate feedback: the original 500 iterations were roughly 500 ms at 1,000 Hz, not the five seconds previously claimed. A wait may overrun by one bounded transfer or poll; setup and cleanup take additional time.

V7 power-on and VCOM preparation return errors through optional checked callbacks and attempt both shutdown writes on failure. Public `epd_poweron_checked()` and `epd_poweroff_checked()` let applications skip drawing or recover when a transition fails. Other boards return `ESP_ERR_NOT_SUPPORTED` without changing power. Existing void entry points remain available. If a bus fault prevents cleanup reaching the expander, that failure is logged separately; physical shutdown cannot be guaranteed by software.

Validation: `python3 test/host/test_power.py` compiles the actual power functions and I2C helpers with sanitizers at 100 Hz and 1,000 Hz. It covers healthy/delayed startup, both missing signals, stalled I2C, late readiness, each transfer failing during normal/Carta/VCOM startup, shutdown failures, and legacy/unsupported API behaviour. An ESPHome application using the checked APIs builds successfully for the LilyGO T5 E-Paper S3 Pro with ESP-IDF 5.5.4. Hardware fault cases are simulated, not deliberately induced on a live panel.

### DIFF
--- a/src/board/epd_board_v7.c
+++ b/src/board/epd_board_v7.c
@@ -228,7 +228,14 @@ static void epd_board_poweron(epd_ctrl_state_t* state) {
     // give the IC time to powerup and set lines
     vTaskDelay(1);
 
+    int pg_tries = 0;
     while (!(pca9555_read_input(config_reg.i2c.pca, 1) & CFG_PIN_PWRGOOD)) {
+        if (pg_tries >= 500) {
+            ESP_LOGE("epdiy", "Power enable failed! PWRGOOD (PCA9555) not asserted");
+            return;
+        }
+        pg_tries++;
+        vTaskDelay(1);
     }
 
     ESP_ERROR_CHECK(tps_write_register(config_reg.i2c.tps, TPS_REG_ENABLE, 0x3F));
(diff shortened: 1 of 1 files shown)


vroland:
Makes sense, thanks! Small nitpick: Can you extract the number of retries into a module-level define? Also, depending on the configured FreeRTOS tick rate the timeout may be longer or shorter. The smallest is 1ms giving us ~500ms which should be fine, but if there is a trivial way to make it tickrate-independent we may want to do that.

martinberlin:
You could simply use something like:

     vTaskDelay(pdMS_TO_TICKS(10)); // 10 milliseconds 
     
 And just loop accordingly to make it 5 seconds (Which is a lot) anyways my feedback about this would be:
 
 If PWR_GOOD signal does not come, most probably your display won't update anyways since the power lines will be not powered. So with timeout or without it, will inevitably fail.

